PR #19, opened 2020-06-15
Fix mypy error: unused 'type: ignore comment

jbueltemeier:
underlying issue was fixed in [pmeier/pystiche/pull/317](https://github.com/pmeier/pystiche/pull/317)


pmeier:
@jbueltemeier Could you fix these errors for the whole code base?

PR #24, opened 2020-06-19
replace TravisCI and AppVeyor with Github Actions

pmeier:


pmeier:
AppVeyor fails are expected, since there is no config left.

PR #35, opened 2020-06-23
Add tests for gatys_et_al_2017 utils

jbueltemeier:


tests/test_gatys_et_al_2017.py
```diff
@@ -23,3 +26,20 @@ def test_gatys_et_al_2017_images(subtests):
 
 def test_gatys_et_al_2017_preprocessor():
     assert isinstance(utils.gatys_et_al_2017_preprocessor(), CaffePreprocessing)
+
+
+def test_gatys_et_al_2017_postprocessor():
+    assert isinstance(utils.gatys_et_al_2017_postprocessor(), CaffePostprocessing)
+
+
+def test_gatys_et_al_2017_multi_layer_encoder():
```
pmeier, line 35:
```suggestion
@pytest.mark.large_download
@pytest.mark.slow
def test_gatys_et_al_2017_multi_layer_encoder():
```

Without these marks you would try to download the pre-trained weights, which is not only slow but also encumbering for the CI. Later we should mock the download to be able to still test this. I'' take that up

pmeier:
@jbueltemeier I went ahead and fixed the merge conflicts.

PR #37, opened 2020-06-23
Add test for gatys_et_al_2017 content_loss

jbueltemeier:


tests/gatys_et_al_2017/test_loss.py
```diff
@@ -0,0 +1,28 @@
...
+from pystiche_papers.gatys_et_al_2017 import loss
+
+
+@pytest.fixture
+def setup(multi_layer_encoder_with_layer, target_image, input_image):
+    multi_layer_encoder, layer = multi_layer_encoder_with_layer
+    encoder = multi_layer_encoder.extract_encoder(layer)
+    target_enc = encoder(target_image)
+    input_enc = encoder(input_image)
+    return multi_layer_encoder, layer, target_enc, input_enc
```
pmeier, line 14: IMO a fixture only makes sense if we are using it for more than one test. Thoughts?

jbueltemeier: We also use the same code in other tests, such as this [test_loss.py](https://github.com/pmeier/pystiche_papers/blob/master/tests/gatys_ecker_bethge_2015/test_loss.py) in gatys_ecker_bethge_2015. The `target_enc` and `input_enc` are also needed in further tests where a `MultiLayerEncoder` is used.

pmeier: Yeah, you are right about that. Still, we cant do that here. We should think about a better concept for this. I'll inline this code for now.

PR #61, opened 2020-07-13
add workflow to test if publishable

pmeier:


pyproject.toml
```diff
@@ -55,6 +56,11 @@ exclude = '''
...
   | \.tox
   | __pycache__
+  | build
   | docs/build
 )/
 '''
+
+[tool.check-wheel-contents]
+# See https://github.com/jwodder/check-wheel-contents/issues/2
+ignore = "W002"
```
pmeier, line 66: Should be removed when https://github.com/jwodder/check-wheel-contents/issues/2 is resolved.

PR #66, opened 2020-07-20
Add test for gatys_et_al_2017 nst

jbueltemeier:


tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,50 @@
...
+    get_optimizer = kwargs["get_optimizer"]
+    preprocessor = kwargs["preprocessor"]
+    postprocessor = kwargs["postprocessor"]
+
+    with subtests.test("input_image"):
+        ptu.assert_allclose(input_image, resize(content_image, 800))
+
+    with subtests.test("criterion"):
+        assert type(criterion) is type(  # noqa: E721
+            paper.gatys_et_al_2017_perceptual_loss()
```
pmeier, line 27:
```suggestion
        assert isinstance(criterion, type(paper.gatys_et_al_2017_perceptual_loss()))
```
I think that is more _pythonic_. Could you also address this below and in a follow-up PR also in https://github.com/pmeier/pystiche_papers/blob/aba7a3d55bc2697f52da78caffc79936a13cbe31/tests/gatys_ecker_bethge_2015/test_core.py#L31-L32 ?

tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+
+    with subtests.test("pyramid"):
+        assert type(pyramid) is type(  # noqa: E721
+            paper.gatys_et_al_2017_image_pyramid()
+        )
+
+    with subtests.test("optimizer"):
+        assert hasattr(get_optimizer, "__call__")
```
pmeier, line 36: Could you add a `is_callable` function to `tests/_utils.py` and use that here? I think that would make it more clear. Please also address this in a follow-up PR in https://github.com/pmeier/pystiche_papers/blob/aba7a3d55bc2697f52da78caffc79936a13cbe31/tests/gatys_ecker_bethge_2015/test_core.py#L36

tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,50 @@
...
+    paper.gatys_et_al_2017_nst(content_image, style_image)
+
+    args, kwargs = mock.call_args
+    input_image, criterion, pyramid = args
+    get_optimizer = kwargs["get_optimizer"]
+    preprocessor = kwargs["preprocessor"]
+    postprocessor = kwargs["postprocessor"]
+
+    with subtests.test("input_image"):
+        ptu.assert_allclose(input_image, resize(content_image, 800))
```
pmeier, line 23: We should use `pyramid[-1].resize_image(content_image)` here. If we are sure the `pyramid` is working correctly we don't need to know the actual size of the image.

tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,47 @@
...
+    with subtests.test("input_image"):
+        ptu.assert_allclose(input_image, pyramid[-1].resize_image(content_image))
+
+    with subtests.test("criterion"):
+        assert isinstance(criterion, type(paper.gatys_et_al_2017_perceptual_loss()))
+
+    with subtests.test("pyramid"):
+        assert type(pyramid) is type(  # noqa: E721
+            paper.gatys_et_al_2017_image_pyramid()
+        )
```
pmeier, line 30: Please use `isinstance` as above.

tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,47 @@
...
+
+    with subtests.test("optimizer"):
+        assert is_callable(get_optimizer)
+        optimizer = get_optimizer(input_image)
+        assert type(optimizer) is type(  # noqa: E721
+            utils.gatys_et_al_2017_optimizer(input_image)
+        )
```
pmeier, line 37: Please use `isinstance` as above.

tests/gatys_et_al_2017/test_core.py
```diff
@@ -0,0 +1,47 @@
...
+
+    with subtests.test("preprocessor"):
+        assert type(preprocessor) is type(  # noqa: E721
+            utils.gatys_et_al_2017_preprocessor()
+        )
+
+    with subtests.test("postprocessor"):
+        assert type(postprocessor) is type(  # noqa: E721
+            utils.gatys_et_al_2017_postprocessor()
+        )
```
pmeier, line 47: Please use `isinstance` as above.

PR #69, opened 2020-07-21
Tests gatys et al 2017 guided loss

jbueltemeier:


tests/_fixtures.py
```diff
@@ -97,6 +100,34 @@ def style_image(image_small_2):
...
+                torch.ones(
+                    [1, 1, ceil(target_image.size()[2] / 2), target_image.size()[3]]
+                ),
+                torch.zeros(
+                    [1, 1, ceil(target_image.size()[2] / 2), target_image.size()[3]]
+                ),
+            ),
+            2,
+        ),
+        "botttom": torch.stack(
```
pmeier, line 117:
```suggestion
        "bottom": torch.stack(
```
typo

tests/_fixtures.py
```diff
@@ -97,6 +100,34 @@ def style_image(image_small_2):
     return image_small_2
 
 
+@pytest.fixture
+def guides(target_image):
+    return {
```
pmeier, line 105: It would be better to create a bool "guide" first and than invert it for the second region. With this you can be sure that the whole canvas is filled and the regions do not overlap.

tests/_fixtures.py
```diff
@@ -97,6 +100,34 @@ def style_image(image_small_2):
     return image_small_2
 
 
+@pytest.fixture
+def guides(target_image):
```
pmeier, line 104: I think we should create a function `create_guides` in `_utils` that does this and use that in a fixture called `input_guides` and `target_guides`. With this we are flexible to make guides for other sized images as well.

tests/_fixtures.py
```diff
@@ -23,6 +24,8 @@
     "input_image",
     "content_image",
     "style_image",
+    "content_guides",
```
pmeier, line 27: Could you add fixtures for `target_guides` and `input_guides`?

tests/_utils.py
```diff
@@ -63,3 +64,18 @@ def load_module(location):
...
+            torch.ones(
+                [1, 1, ceil(image.size()[2] / 2), image.size()[3]], dtype=torch.bool
+            ),
+            torch.zeros(
+                [1, 1, ceil(image.size()[2] / 2), image.size()[3]], dtype=torch.bool
+            ),
+        ),
+        2,
+    )
+    return {"top": mask, "bottom": ~mask}
```
pmeier, line 81: `pystiche` expects guides to be `torch.flaot32`:

```suggestion
    return {"top": mask.float(), "bottom": ~mask.float()}
```

tests/_utils.py
```diff
@@ -63,3 +64,18 @@ def load_module(location):
 
 def is_callable(obj):
     return hasattr(obj, "__call__")
+
+
+def create_guides(image):
+    mask = torch.stack(
```
pmeier, line 70: For more clarity I would do 

```python
from pystiche import image

height, width = image.extract_image_size(image)
```

tests/_utils.py
```diff
@@ -63,3 +64,18 @@ def load_module(location):
...
+        (
+            torch.ones(
+                [1, 1, ceil(image.size()[2] / 2), image.size()[3]], dtype=torch.bool
```
pmeier, line 73: If the image height is odd the guide size will mismatch since it is rounded up twice. I would do

```python
top_height = height // 2 # built in floor
bottom_height = height - top_height
```

This way you are safe

tests/gatys_et_al_2017/test_loss.py
```diff
@@ -80,6 +80,28 @@ def test_gatys_et_al_2017_style_loss(subtests):
         assert style_loss.score_weight == pytest.approx(1e3)
 
 
+def gatys_et_al_2017_guided_style_loss(subtests, content_guides):
```
pmeier, line 83: I didn't see this before, but if I'm not mistaken we never use the actual guides. Couldn't we do this at the beginning of the test?

```python
regions = tuple(f"region{idx}" for idx in range(3))
```

and test everything with that? IMO we don't need a fixture for that.

jbueltemeier: Here we do not need the actual guides yet. However, we will need them later for `guided_nst`. Therefore I wanted to introduce the guides here to use these fixtures consistently for the guides.

PR #71, opened 2020-07-21
add load_weights mock to gatys_et_al_2017 tests

pmeier:
Without this every (indirect) call to `gatys_et_al_2017_multi_layer_encoder` will load the weights. This significantly slows down the test suite.


pmeier:
To prove my point:

[Without this commit](https://github.com/pmeier/pystiche_papers/runs/893278512): 50s
[With this commit](https://github.com/pmeier/pystiche_papers/pull/71/checks?check_run_id=894209168): 28s

PR #80, opened 2020-07-22
Add test li_wand_2016 pyramid

jbueltemeier:


tests/li_wand_2016/test_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_li_wand_2016_image_pyramid(subtests, mocker):
+
+    configs = ((True, 100, 3), (False, 200, None))
+    for (impl_params, num_steps, num_levels,) in configs:
+        with subtests.test(impl_params=impl_params):
+
+            if num_levels is None:
+                num_levels = int(np.floor(np.log2(6))) + 1
```
pmeier, line 13: What is this for? If this is just to get around `num_levels = None`, see below.
```suggestion
```

tests/li_wand_2016/test_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+                pyramid_edge = kwargs["edge"]
+
+                with subtests.test("max_edge_size"):
+                    assert pyramid_max_edge_size == 384
+
+                with subtests.test("num_steps"):
+                    assert pyramid_num_steps == num_steps
+
+                with subtests.test("num_levels"):
+                    assert pyramid_num_levels == num_levels
```
pmeier, line 31: If we don't give a specific value for `num_levels` we can trust the `OctaveImagePyramid` to handle it.

```suggestion
                    if isinstance(num_levels, int):
                        assert pyramid_num_levels == num_levels
                    else:
                        assert isinstance(pyramid_num_levels, int)
```

Issue #20, opened 2020-04-01
RuntimeError during texture synthesis


jbueltemeier:
An `RuntimeError` occurs when trying to synthesize a texture with a `ulyanov_et_al_2016_transformer`, i. e. when `sytlization=False` and for example `input=(256,256)`. 

```python
from pystiche.papers.ulyanov_et_al_2016.core import ulyanov_et_al_2016_texture_synthesis
from pystiche.papers.ulyanov_et_al_2016.modules import ulyanov_et_al_2016_transformer
transformer = ulyanov_et_al_2016_transformer(stylization=False).to("cuda")
ulyanov_et_al_2016_texture_synthesis((256,256), transformer, stylization=False)
```

pmeier:
Since #129 removed the texture synthesis functionality, this error can no longer happen. As such I'm closing this.

PR #85, opened 2020-07-27
Tests ulyanov 2016 loss

jbueltemeier:


tests/ulyanov_et_al_2016/test_loss.py
```diff
@@ -1,32 +1,155 @@
...
+        with subtests.test(impl_params=impl_params):
+            op = loss.UlyanovEtAl2016FeatureReconstructionOperator(
+                encoder, impl_params=impl_params,
+            )
+            op.set_target_image(target_image)
+            actual = op(input_image)
+
+            score = mse_loss(input_enc, target_enc)
+
+            desired = score / extract_batch_size(input_enc) if impl_params else score
```
pmeier, line 32: By default the `input_image` and `target_image` have `batch_size==1`. Thus if we don't increase it manually, this will work regardless of a "correct" implementation.

tests/ulyanov_et_al_2016/test_loss.py
```diff
@@ -1,32 +1,155 @@
...
+                    score_weight = 1e0
+                else:
+                    score_weight = 1e3 if stylization else 1e0
+            else:
+                score_weight = 1e0
+            assert style_loss.score_weight == pytest.approx(score_weight)
+
+
+def test_ulyanov_et_al_2016_perceptual_loss(subtests):
+    for stylization in (True, False):
```
pmeier, line 136: For now you can ignore all cases with `stylization is False`. We probably want to change the API since it leads to errors such as #20. The discussion of this is out of the scope of this PR though.

tests/ulyanov_et_al_2016/test_loss.py
```diff
@@ -1,32 +1,147 @@
...
+from pystiche.loss import PerceptualLoss
 from pystiche_papers.ulyanov_et_al_2016 import loss
 
 
-def test_ulyanov_et_al_2016_content_loss(
-    subtests, multi_layer_encoder_with_layer, batch_image
+def test_UlyanovEtAl2016FeatureReconstructionOperator(
+    subtests, multi_layer_encoder_with_layer, target_image, input_image
 ):
+    input_image = torch.cat((input_image, input_image), 0)
```
pmeier, line 18: Nit 
Nit

```suggestion
    from pystiche_papers.utils import batch_up_image
    input_image = batch_up_image(input_image, 2)
```

tests/ulyanov_et_al_2016/test_loss.py
```diff
@@ -1,32 +1,147 @@
...
     multi_layer_encoder, layer = multi_layer_encoder_with_layer
     encoder = multi_layer_encoder.extract_encoder(layer)
-    target_image = batch_image
-    input_image = batch_image.flip((0,))
     target_enc = encoder(target_image)
```
pmeier, line 21: I think we should do the same with `target_image` and another `batch_size` so we can be sure that it has no effect.

jbueltemeier: What do you mean by different `batch_size`? Due to the `mse_loss` the `batch _size` of the `target_image` can only have the size 1 or the same size as the `input_image`.

pmeier: Ah sorry, I thought we were testing the style loss, where the `batch_size` is irrelevant :facepalm:. In that case, go for the same `batch_size` as the `input_image`.

Issue #87, opened 2020-07-27
Exclude texture synthesis from package scope


pmeier:
Currently we switch between NST and texture synthesis (TS) with the `stylization` flag for the Ulyanov et al. paper. Although the implementation is quite similar, the input is different. This leads to bugs (#20) with quirky workarounds. 

On one hand we could have two separate models (`transformer` and `synthesizer`) for both use cases, which would make the interface cleaner. On the other hand I'm considering to remove the TS capability all together, since `pystiche_papers` was created to replicate NST papers.

I'm open for a discussion about this. Cc @jbueltemeier

jbueltemeier:
I would think that this functionality can be removed because, as you said, `pystiche_papers` was created for replication of NST papers and not for TS. It would only be a nice additional functionality with which you could show that TS can also be performed with the `pystiche` framework. However, since you don’t expect this additional functionality and this function requires an extra effort, I don’t see a reason to put more time in it to run it without bugs.

pmeier:
I'll take it up if we have sufficient coverage for the NST part.

PR #89, opened 2020-07-28
Tests ulyanov 2016 basic modules

jbueltemeier:


tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
+import itertools
+from typing import Dict, cast
```
pmeier, line 2: Static typing is not needed nor checked in the test suite.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
...
+        ptu.assert_allclose(join_image[:, :input_num_channels, :, :], input_image)
+    with subtests.test("style_image"):
+        ptu.assert_allclose(join_image[:, input_num_channels:, :, :], style_image)
+
+
+def test_UlyanovEtAl2016StylizationDownsample(subtests):
+    module = modules.UlyanovEtAl2016StylizationDownsample()
+    assert isinstance(module, nn.AvgPool2d)
+    with subtests.test("kernel_size"):
+        assert module.kernel_size == pytest.approx(2)
```
pmeier, line 60: `kernel_size` should be an `int` or a `tuple` of `int`s. Thus, `pytest.approx` is not needed here.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
...
+                assert norm_module.eps == pytest.approx(1e-5)
+
+            with subtests.test("momentum"):
+                assert norm_module.momentum == pytest.approx(1e-1)
+
+            with subtests.test("affine"):
+                assert norm_module.affine
+
+            with subtests.test("track_running_stats"):
+                assert norm_module.track_running_stats
```
pmeier, line 123: We don't need to test this since it has no effect on the behavior.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def test_get_activation_module(subtests):
+    for instance_norm, impl_params in itertools.product((True, False), (True, False)):
```
pmeier, line 127:
```suggestion
    for impl_params, instance_norm in itertools.product((True, False), (True, False)):
```

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
...
+        with subtests.test(instance_norm=instance_norm):
```
pmeier, line 128:
```suggestion
        with subtests.test(impl_params=impl_params, instance_norm=instance_norm):
```

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,143 @@
...
+
+            assert isinstance(
+                type(norm_module),
+                type(nn.ReLU) if impl_params and instance_norm else type(nn.LeakyReLU),
+            )
+
+            with subtests.test("inplace"):
+                assert norm_module.inplace
+
+            if not impl_params and instance_norm:
```
pmeier, line 141: At this point we have already checked if we only get an `nn.LeakyReLU` for the correct parameter combination.

```suggestion
            if isinstance(norm_module, nn.LeakyReLU):
```

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,136 @@
...
+from pystiche_papers.ulyanov_et_al_2016 import modules
+
+
+def test_SequentialWithOutChannels(subtests):
+    sequential_modules = [nn.Conv2d(3, 3, 1), nn.Conv2d(3, 3, 1)]
+    for out_channel_name in (
+        None,
+        1,
+        "1",
+    ):  # TODO: Rename out_channel_name is not implemented
```
pmeier, line 19: Could you clarify what you mean by this?

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,136 @@
...
+                desired_out_channel_name = "1"
+            elif isinstance(out_channel_name, int):
+                desired_out_channel_name = str(out_channel_name)
+            else:
+                desired_out_channel_name = out_channel_name
+
+            assert tuple(module._modules.keys())[-1] == desired_out_channel_name
+
+        with subtests.test("out_channels"):
+            assert module.out_channels == sequential_modules[-1].out_channels
```
pmeier, line 35: I think you misunderstood the purpose of this class. I've introduced it so that we have access to an `.out_channels` attribute on an `nn.Sequential`. Thus, you should test if this attribute matches the expected `out_channels`. On a side note: we should probably move this to `pystiche_papers.utils.modules` in a follow-up PR.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def test_join_channelwise(subtests, input_image, style_image):
```
pmeier, line 38: I would prefer you use `image_small_0` and `image_small_1` since this has nothing to do with the `style_image`.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,136 @@
...
+    join_image = modules.join_channelwise(input_image, style_image)
+    assert isinstance(join_image, torch.Tensor)
+    input_num_channels = extract_num_channels(input_image)
+    with subtests.test("num_channels"):
+        assert extract_num_channels(
+            join_image
+        ) == input_num_channels + extract_num_channels(style_image)
```
pmeier, line 45: This does not need to be a subtest.

tests/ulyanov_et_al_2016/test_modules.py
```diff
@@ -0,0 +1,136 @@
...
+    mock = mocker.patch(
+        "pystiche_papers.ulyanov_et_al_2016.modules.TextureNoiseParams.downsample"
+    )
+    module = modules.UlyanovEtAl2016TextureDownsample()
+    module(input_image)
+    mock.assert_called_once()
+
+
+def test_ulyanov_et_al_2016_downsample(subtests):
+    for stylization in (True, False):
```
pmeier, line 73: See comment from previous review or discussion in #87.

pmeier:
@jbueltemeier that merge was a little early :wink:. Please address the comments in a follow-up PR.

tests/ulyanov_et_al_2016/test_modules.py
jbueltemeier: Currently, any input other than this will result in a `KeyError`, because the key will not be renamed.

tests/ulyanov_et_al_2016/test_modules.py
jbueltemeier: I use this query to check if the `out_channels` of the `SequentialWithOutChannels` match the `out_channels` of the last entered module in `sequential_modules`, which are the expected ones. Or do I misunderstand the remark?

pmeier: My bad, I misunderstood L35.

PR #92, opened 2020-07-28
Remove double name spacing and limited exposure of reference implementations

### LINKED ISSUE #56: Double name spacing
Currently, every paper reference implementation is name-spaced twice. For Example:

```python
pystiche_papers.gatys_ecker_bethge_2015.gatys_ecker_bethge_nst()
```

This is very cumbersome and does not provide a lot in return. The only time I can think of where this would be helpful if one imports it at the top of a file and only uses it far away from the imports:

```python
from pystiche_papers.gatys_ecker_bethge_2015 import gatys_ecker_bethge_nst

a_lot_of_lines_since_top_of_file = True

output_image = gatys_ecker_bethge_nst()
```

---

I suggest we remove the second name-spacing, i. e. 

```python
pystiche_papers.gatys_ecker_bethge_2015.nst()
```

If we import the sub-package opposed to individual functions we also do not lose explicitness:

```python
from pystiche_papers import gatys_ecker_bethge_2015

a_lot_of_lines_since_top_of_file = True

output_image = gatys_ecker_bethge.nst()
```

Furthermore this allows us to do 

```python
from pystiche_papers import gatys_ecker_bethge_2015 as paper

paper.nst()
```
which is more concise if the context allows it.

@jbueltemeier If we want to do this, we should wait until we reach significant test coverage in order not to break anything.

### LINKED ISSUE #57: Expose most functions and classes of the reference implementations
Currently only few functions for each reference implementation are directly exposed:

https://github.com/pmeier/pystiche_papers/blob/5edfd80d12a038a0ef663f992aa3fc8e87ed8d2c/pystiche_papers/gatys_ecker_bethge_2015/core.py#L18-L22

This can result in requiring multiple imports for no apparent reason:

https://github.com/pmeier/pystiche_papers/blob/5edfd80d12a038a0ef663f992aa3fc8e87ed8d2c/tests/gatys_ecker_bethge_2015/test_core.py#L8-L9

I think we should simply expose everything that is already name-spaced, i.e. prefixed by `author1_author_2_year` or `Author1Author2Year`.

Cc @jbueltemeier

pmeier:
Closes #56, closes #57.


pmeier:
I'll add the changes for `johnson_alahi_li_2016` and `ulyanov_et_al_2016` when we finished the tests.

PR #94, opened 2020-07-29
Tests ulyanov 2016 data

jbueltemeier:


tests/ulyanov_et_al_2016/test_data.py
```diff
@@ -1,20 +1,143 @@
...
+                    make_reproducible()
+                    transform = ValidRandomCrop(edge_size)
+                    transform_image = transform(image)
+                else:
+                    transform_image = resize(image, edge_size)
+            else:
+                transform = CenterCrop(edge_size)
+                transform_image = transform(image)
+
+            desired = grayscale_to_fakegrayscale(transform_image)
```
pmeier, line 46: I would split this off into a separate test, which test for this explicitly. The standard case should be RGB.

tests/ulyanov_et_al_2016/test_data.py
```diff
@@ -1,20 +1,143 @@
...
+
+            ptu.assert_allclose(actual, desired)
+
+
+def test_ulyanov_et_al_2016_style_transform(subtests):
```
pmeier, line 51: This is OK for now, but we really should refactor

https://github.com/pmeier/pystiche_papers/blob/f77f1162846519a85b33932627a0dfbd57bf35f8/pystiche_papers/ulyanov_et_al_2016/data.py#L62-L73

1. `edge_size`s default should simply be `256`
2. There is no need for a `ComposedTransform` if we only use a single `Resize`.

Could you address this in a follow-up PR?

tests/ulyanov_et_al_2016/test_data.py
```diff
@@ -1,20 +1,143 @@
...
+            assert isinstance(batch_sampler, FiniteCycleBatchSampler)
+
+            with subtests.test("num_batches"):
+                assert (
+                    batch_sampler.num_batches == 200
+                    if not impl_params
+                    else 2000
+                    if instance_norm
+                    else 300
+                )
```
pmeier, line 121: This is quite hard to read. Could you use nested `if`s here?

tests/ulyanov_et_al_2016/test_data.py
```diff
@@ -1,20 +1,143 @@
...
+
+            with subtests.test("num_batches"):
+                assert (
+                    batch_sampler.batch_size == 16
+                    if not impl_params
+                    else 1
+                    if instance_norm
+                    else 4
+                )
```
pmeier, line 130: Same as above.

PR #98, opened 2020-07-29
temporarily split off windows tests

pmeier:
For yet unknown reasons, as of this morning (29 July 2020) `pip install` requires the `--user` flag when running on a Windows machine in GitHub actions. I've opened a report [here](https://github.community/t/pip-install-suddenly-requires-user-flag-on-windows/125056). 

This splits the workflow in `linux-macos` and `windows` to account for the differences. This should be reverted when the behavior change in GitHub Actions is resolved.

Test failure is expected and will be handled in #97.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,10 +20,10 @@ on:
       - "tox.ini"
 
 jobs:
-  unit:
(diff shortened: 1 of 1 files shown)


pmeier:
superseded by #99

PR #106, opened 2020-07-31
enable PyTorch hub

pmeier:


pmeier:
Upcoming commits will fail, since we switch the test branch to `master`. This will only work after this PR is merged.

PR #111, opened 2020-08-03
enable hub tests from CI

pmeier:
This adds the environment variable `PYSTICHE_HUB_GITHUB` which can be used to override the default `pmeier/pystiche_papers:master`. With this hub testing should be possible from PRs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,11 +59,7 @@ jobs:
         run: tox --notest
 
       - name: Run tests
-        env:
-          PYSTICHE_HUB_GITHUB: "$GITHUB_ACTOR/pystiche_papers:$GITHUB_HEAD_REF"
-        run: |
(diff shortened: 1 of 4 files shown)


pmeier:
This should also work for forks, but I'm not sure. @jbueltemeier Could you create a file `$PYSTICHE_PAPERS_ROOT/.github/workflows/gha-env-vars.yml` with the content below and open a PR with it?

```yml
name: tests

on:
  pull_request:

jobs:
  test:
    runs-on: ubuntu-latest

    steps:
      - name: Environment variables
        run: |
          echo GITHUB_ACTOR $GITHUB_ACTOR
          echo GITHUB_REPOSITORY $GITHUB_REPOSITORY
          echo GITHUB_REF $GITHUB_REF
          echo GITHUB_BASE_REF $GITHUB_BASE_REF
          echo GITHUB_HEAD_REF $GITHUB_HEAD_REF

```

PR #118, opened 2020-08-04
Add doc for johnson_alahi_li_2016

jbueltemeier:


pmeier:
@jbueltemeier Please rebase from the current master. In #122 and #123 I've added the framework to "see the doc strings in action".

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
     layer: str = "relu2_2",
     score_weight: Optional[float] = None,
 ) -> ops.FeatureReconstructionOperator:
+    r"""Content_loss from "Perceptual Losses for Real-Time Style Transfer and
+    Super-Resolution" by Johnson, Alahi, and Li (original authors) :cite:`JAL2016`.
```
pmeier, line 42: I wonder if we can shorten this somehow. I don't think we need this block at the top of every docstring. Maybe "Content loss from :cite:`JAL2016`"?

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
```
pmeier, line 47: Where do we want to put these details? We could either gather them in one point (for example above the API in the docs) or within each docstring. Thoughts?

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+        style: Optional style for selecting the correct optimization parameters for the reproduction.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.multi_layer_encoder.MultiLayerEncoder`. If ``None``, the
```
pmeier, line 51:
```suggestion
        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If ``None``, the
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+            default :func:`~pystiche_papers.johnson_alahi_li_2016._utils.multi_layer_encoder` from the paper is used.
```
pmeier, line 52: You can use these references just like an `import`

```suggestion
            default :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder` from the paper is used.
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
pmeier, line 51:
```suggestion
        multi_layer_encoder: Pretrained :class:`~pystiche.enc.multi_layer_encoder.MultiLayerEncoder`. If omitted, the
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+            Defaults to ``None``.
```
pmeier, line 53:
```suggestion
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+        layer: Layer from which the encs of the :class:`~pystiche.enc.multi_layer_encoder.MultiLayerEncoder`
```
pmeier, line 54:
```suggestion
        layer: Layer from which the encodings of the ``multi_layer_encoder``
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+            should be taken. Defaults to "relu2_2".
+        score_weight: Score weight of the operator. If ``None``, the score_weight is determined by the
```
pmeier, line 56:
```suggestion
        score_weight: Score weight of the operator. If omitted, the score_weight is determined with respect to ``style`` and ``impl_params``. For details see FIXME
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+            :func:`get_content_score_weight`. Defaults to ``None``.
```
pmeier, line 57: This function is not considered public.

```suggestion
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -38,6 +38,25 @@ def content_loss(
...
+    r"""Content_loss from "Perceptual Losses for Real-Time Style Transfer and
+    Super-Resolution" by Johnson, Alahi, and Li (original authors) :cite:`JAL2016`.
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+        style: Optional style for selecting the correct optimization parameters for the reproduction.
```
pmeier, line 50: 1. What do you mean by correct? 
2. I would swap "reproduction" with "replication"

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -262,6 +292,7 @@ def __init__(
             self.init_weights()
 
     def init_weights(self) -> None:
+        r"""Initializes the weights of the transformer as described in the paper."""
```
pmeier, line 295: This is not described in the paper, but rather implemented like this.

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,35 @@ def training(
...
+        quiet: If ``True``, not information is logged during the optimization. Defaults
+            to ``False``.
+        logger: Optional custom logger. If ``None``,
+            :class:`pystiche.optim.OptimLogger` is used. Defaults to ``None``.
+        log_fn: Optional custom logging function. It is called in every optimization
+            step with the current step and loss. If ``None``,
+            :func:`~pystiche.optim.default_image_optim_log_fn` is used. Defaults to
+            ``None``.
+
+    Returns: Trained transformer for the NST.
```
pmeier, line 64:
```suggestion
```

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -100,6 +129,26 @@ def stylization(
...
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. If ``None``,
+            ``instance_norm`` is set to ``impl_params``. Defaults to ``None``.
+        framework: Framework that was used to train the the transformer. Can be one of
+            ``"pystiche"`` (default) and ``"luatorch"``.
+        preprocessor: Optional preprocessor that is called with the ``input_image``
+            before the optimization.
+        postprocessor: Optional preprocessor that is called with the ``output_image``
+            after the optimization.
+
+    Returns: Stylised ``input_image``.
```
pmeier, line 150:
```suggestion
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
jbueltemeier: I think you are right and for the docstrings this shortened form is sufficient. But is it also possible to integrate the complete form with links to the original paper and the original github repository above the single docstrings for pystiche_papers.johnson_alahi_li_2016 in general?

pmeier: Sure. I'll add an example for this later.

pystiche_papers/johnson_alahi_li_2016/_loss.py
jbueltemeier: I think above the API is the better choice.

pmeier: This was my first idea, but I disagree with it now. I propose we put a summary ("These functions include changes compared to the paper") on top of the API, but describe the actual changes within the docstring. I wouldn't do this in the `Args` section, but rather below it. Could you make a suggestion?

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -249,6 +265,17 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
 class Transformer(nn.Module):
+    r"""Transformer from :cite:`JAL2016` .
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+        init_weights: If ``True``, use the procedure defined in the :func:`init_weights` to initialize the weights.
```
pmeier, line 276:
```suggestion
        init_weights: If ``True``, use :meth:`init_weights` to initialize the weights the same way the original implementation did.
```

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -262,6 +289,7 @@ def __init__(
             self.init_weights()
 
     def init_weights(self) -> None:
+        r"""Initializes the weights of the transformer as it is the case in the original implementation."""
```
pmeier, line 292: Please describe the initialization. Should be one of "Kaiming" or "He", but I'm not sure.

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -262,6 +289,7 @@ def __init__(
...
         for module in self.modules():
             if isinstance(module, (nn.Conv2d, nn.ConvTranspose2d)):
```
pmeier, line 294: Please add a link to the original luatorch initalization scheme for these layers.

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,34 @@ def training(
         Callable[[int, Union[torch.Tensor, pystiche.LossDict], float, float], None]
     ] = None,
 ) -> nn.Module:
+    r"""Training a transformer for the NST.
+
+    Args:
+        content_image_loader: Content images used as input for the ``transformer``.
+        style_image: Style image on which the ``transformer`` should be trained.
```
pmeier, line 41: What is the difference between passing a `str` and a `torch.Tensor`?

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,34 @@ def training(
...
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. If ``None``,
+            ``instance_norm`` is set to ``impl_params``. Defaults to ``None``.
```
pmeier, line 47:
```suggestion
            :class:`~torch.nn.BatchNorm2d` as described in the paper. If omitted,
            defaults to ``impl_params``.
```

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,34 @@ def training(
...
+        transformer: Transformer to be optimized. If ``None``, the default
+            :func:`~pystiche_papers.johnson_alahi_li_2016.transformer` from the paper is used.
```
pmeier, line 49:
```suggestion
        transformer: Transformer to be optimized. If omitted, the
            :func:`~pystiche_papers.johnson_alahi_li_2016.transformer` is used.
```

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -100,6 +128,25 @@ def stylization(
     preprocessor: Optional[nn.Module] = None,
     postprocessor: Optional[nn.Module] = None,
 ) -> torch.Tensor:
+    r"""Transforms an input image into a stylised version using the transfromer.
+
+    Args:
+        input_image: Image to be stylised.
+        transformer: Pretrained transformer for style transfer or string to load a pretrained transformer.
```
pmeier, line 135: This string is the `style` parameter of `transformer()`.

pystiche_papers/johnson_alahi_li_2016/_utils.py
```diff
@@ -15,10 +15,27 @@ def postprocessor() -> transforms.CaffePostprocessing:
 
 
 def multi_layer_encoder(impl_params: bool = True,) -> enc.VGGMultiLayerEncoder:
+    r"""Multi-layer encoder based on the VGG16 architecture with the weights of caffe,
```
pmeier, line 18: Is there a reason you diverted from the scheme to simply put the name and citation in the first docstring line?

pmeier:
Test failure is unrelated. www.wassilykandinsky.net seems to be down.

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -54,6 +54,27 @@ def content_loss(
     layer: str = "relu2_2",
     score_weight: Optional[float] = None,
 ) -> ops.FeatureReconstructionOperator:
+    r"""Content_loss from :cite:`JAL2016`.
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
```
pmeier, line 62: Please put a description of what exactly happens below the `Args` section.

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -54,6 +54,27 @@ def content_loss(
...
+            replication of the given pre-trained models from the original repository.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            ``omitted``, the default
+            :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder` from the
+            paper is used.
+        layer: Layer from which the encodings of the ``multi_layer_encoder`` should be
+            taken. Defaults to "relu2_2".
+        score_weight: Score weight of the operator. If ``omitted``, the score_weight is
+            determined with respect to ``style`` and ``instance_norm``. For details see
+            FIXME
```
pmeier, line 75: This should link the table in https://github.com/pmeier/pystiche_papers/blob/master/docs/source/api/pystiche_papers.johnson_alahi_li_2016.rst

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -118,6 +139,30 @@ def style_loss(
...
+    r"""Style_loss from :cite:`JAL2016`.
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+        style: Optional style for selecting the optimization parameters for the
+            replication of the given pre-trained models from the original repository.
```
pmeier, line 151: "See score_weight for details"

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -118,6 +139,30 @@ def style_loss(
...
+            :func:`~pystiche_papers.johnson_alahi_li_2016._utils.multi_layer_encoder`
+            from the paper is used.
+        layers: Layers from which the encodings of the ``multi_layer_encoder`` should be
+            taken. If ``None``, the defaults is used. Defaults to
+            ``("relu1_2", "relu2_2", "relu3_3", "relu4_3")``.
+        layer_weights: Layer weights of the operator. Defaults to "sum".
+        score_weight: Score weight of the operator. If ``omitted``, the score_weight is
+            determined with respect to ``style``, ``instance_norm`` and ``impl_params``.
+            For details see FIXME
+        **gram_op_kwargs: Optional parameters for the
```
pmeier, line 163:
```suggestion
        **gram_op_kwargs: Optional parameters for a
```

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -202,6 +208,19 @@ def encoder(instance_norm: bool = True,) -> pystiche.SequentialModule:
...
+    r"""Decoder part of the Transformer from :cite:`JAL2016` .
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+
+    If ``impl_params`` is ``True``, a tanh with a constant factor of 150 is used instead
```
pmeier, line 220: It is not clear what constant factor means.

1. 150 * tanh(x)
2. tanh(150 * x)

Also

```suggestion
    If ``impl_params is True``, a :func:`~torch.tanh` with a constant factor of 150 is used instead
```

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -202,6 +208,19 @@ def encoder(instance_norm: bool = True,) -> pystiche.SequentialModule:
...
+    of the (tanh(x) + 1) / 2 in the paper.
```
pmeier, line 221: In the paper they use a tanh to scale to [0, 255]. We need to explain why we dropped the 255 constant.

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -263,14 +294,25 @@ def __init__(
             self.init_weights()
 
     def init_weights(self) -> None:
+        r"""Initialize transformer weights.
+
+        Uses the default Kaiming initalisation for uniform distributions in luatorch for
+        :class:`~torch.nn.Conv2d` and :class:`~torch.nn.ConvTranspose2d`. See
+        https://github.com/Kaixhin/nninit#nninitkaimingmodule-tensor-dist-gain .
```
pmeier, line 301: Is this the official source?

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,40 @@ def training(
...
+            :func:`~pystiche_papers.johnson_alahi_li_2016._images`.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. If omitted,
+            defaults to ``impl_params``.
+        transformer: Transformer to be optimized. If omitted, the
+            :func:`~pystiche_papers.johnson_alahi_li_2016.transformer` is used.
+        criterion: Optimization criterion. If ``None``, the default
```
pmeier, line 52: Since `None` plays no other role here than a placeholder for an omitted value, we should make this more clear in the description. I think it is more helpful to say "If omitted, x happens" than "If None, x happens. Defaults to None". Please address this in every docstring.

pystiche_papers/johnson_alahi_li_2016/_nst.py
```diff
@@ -34,6 +34,40 @@ def training(
...
+            to ``False``.
+        logger: Optional custom logger. If ``None``,
+            :class:`pystiche.optim.OptimLogger` is used. Defaults to ``None``.
+        log_fn: Optional custom logging function. It is called in every optimization
+            step with the current step and loss. If ``None``,
+            :func:`~pystiche.optim.default_image_optim_log_fn` is used. Defaults to
+            ``None``.
+
+    If ``impl_params`` is ``True`` , an external instead of an internal preprocessing of
+    the images is used.
```
pmeier, line 68: We should also mention to which functions `impl_params` is passed to.

pystiche_papers/johnson_alahi_li_2016/_utils.py
```diff
@@ -42,10 +42,30 @@ def postprocessor() -> transforms.CaffePostprocessing:
...
 
 def optimizer(transformer: nn.Module) -> optim.Adam:
+    r"""Optimizer from :cite:`JAL2016`.
+
+    Args:
+        transformer: Transformer to be optimized.
+
+    Returns:
+        :class:`torch.optim.Adam` optimizer with a learning rate of ``1e-3``. The
+        parameters of ``transformer`` are set as optimization parameters.
```
pmeier, line 68: The type of the output is added automatically.

```suggestion
        Optimizer for the ``transformer`` parameters.
```

pystiche_papers/johnson_alahi_li_2016/_utils.py
```diff
@@ -42,10 +42,29 @@ def postprocessor() -> transforms.CaffePostprocessing:
...
 
 def multi_layer_encoder(impl_params: bool = True,) -> enc.VGGMultiLayerEncoder:
+    r"""Multi-layer encoder from :cite:`JAL2016`.
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+
+    If ``impl_params`` is ``True`` , an external instead of an internal preprocessing of
```
pmeier, line 52: We can't say anything here about the external preprocessing. We should simply state that no internal preprocessing is used.

pystiche_papers/johnson_alahi_li_2016/_utils.py
```diff
@@ -42,10 +42,29 @@ def postprocessor() -> transforms.CaffePostprocessing:
...
 
 def optimizer(transformer: nn.Module) -> optim.Adam:
+    r"""Optimizer from :cite:`JAL2016`.
+
+    Args:
+        transformer: Transformer to be optimized.
+
+    Returns:
+        Optimizer for the ``transformer`` parameters.
+
```
pmeier, line 68: Although I suggested something else here before, I just feel like this adds no information.

```suggestion
```

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -202,6 +208,21 @@ def encoder(instance_norm: bool = True,) -> pystiche.SequentialModule:
...
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+
+    If ``impl_params is True``, a scaled 150 * :func:`~torch.tanh` is used instead of
+    the (:func:`~torch.tanh`(x) + 1) / 2 in the paper. The scaling in the second variant
+    to [0, 255] is then carried out in the
+    :func:`pystiche_papers.johnson_alahi_li_2016.postprocessor`.
```
pmeier, line 223: We need to talk about that, since its not that simple. Message me when you have time to discuss this in more detail.

docs/source/api/pystiche_papers.johnson_alahi_li_2016.rst
```diff
@@ -1,13 +1,32 @@
...
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.training`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.stylization`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.content_transform`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.style_transform`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.batch_sampler`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.transformer`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.content_loss`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.style_loss`
+  - the :func:`~pystiche_papers.johnson_alahi_li_2016.regularization`
```
pmeier, line 18: Is any of the top-level functions not affected? If not, I think this list is not needed.

docs/source/api/pystiche_papers.johnson_alahi_li_2016.rst
```diff
@@ -1,13 +1,32 @@
...
+
+.. _table-hyperparameters:
```
pmeier, line 20: We probably need to namespace that in case another paper has a similar table.

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -54,6 +54,31 @@ def content_loss(
...
+            :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder` is used.
+        layer: Layer from which the encodings of the ``multi_layer_encoder`` should be
+            taken. Defaults to "relu2_2".
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to ``style`` and ``instance_norm``. For details see
+            :ref:`here <table-hyperparameters>`.
+
+    The parameter ``impl_params`` is passed to the
+    :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder` and a getter
+    function to get the content ``score_weight`` if omitted.
```
pmeier, line 79: I'm probably changing my mind here, but I'm not happy with that as is for two reasons:

1. I (now) think we can omit stating that `impl_params` is passed down. I think that is implicitly clear.
2. The second part does not add any information that is not already described in `score_weight` above.

Thus, I think we can leave it out. Thoughts?

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -54,6 +54,31 @@ def content_loss(
     layer: str = "relu2_2",
     score_weight: Optional[float] = None,
 ) -> ops.FeatureReconstructionOperator:
+    r"""Content_loss from :cite:`JAL2016`.
+
+    Args:
+        impl_params: If ``True``, use the parameters used in the reference
```
pmeier, line 60:
```suggestion
        impl_params: If ``True``, uses the parameters used in the reference
```

pystiche_papers/johnson_alahi_li_2016/_loss.py
```diff
@@ -208,6 +282,27 @@ def perceptual_loss(
...
+            implementation of the original authors rather than what is described in
+            the paper.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper.
+        style: Optional style for selecting the optimization parameters for the
+            replication of the given pre-trained models from the original repository.
+            For details see :ref:`here <table-hyperparameters>`.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`.
```
pmeier, line 296: Default?

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -168,6 +168,15 @@ def maybe_fix_num_channels(num_channels: int, instance_norm: bool) -> int:
...
 
 def encoder(instance_norm: bool = True,) -> pystiche.SequentialModule:
+    r"""Encoder part of the :class:`Transformer` from :cite:`JAL2016` .
+
+    Args:
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. In addition, the
+            number of in_channels and out_channels is reduced to half the number of
+            in_channels and out_channels in each module except for the first
+            in_channels, if this is set to ``True``.
```
pmeier, line 178:
```suggestion
            :class:`~torch.nn.BatchNorm2d` as described in the paper. In addition, the
            number of channels of the convolution layers is reduced by half.
```

pystiche_papers/johnson_alahi_li_2016/_modules.py
```diff
@@ -202,6 +211,26 @@ def encoder(instance_norm: bool = True,) -> pystiche.SequentialModule:
...
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. In addition, the
+            number of in_channels and out_channels is reduced to half the number of
+            in_channels and out_channels in each module except for the last
+            out_channels, if this is set to ``True``.
+
+    If ``impl_params is True``, a scaled 150 * :func:`~torch.tanh` is used instead of
+    the (:func:`~torch.tanh`(x) + 1) / 2 in the paper. Since the reference
+    implementation does not use internal preprocessing for the criterion, the scaled
+    :func:`~torch.tanh` is used, whereby this necessary preprocessing is learned within
+    the transformer. Thus this corresponds to a rather inaccurate scaling to [0, 255].
```
pmeier, line 230:
```suggestion
    If ``impl_params is True``, the output of the decoder is not externally 
    pre-processed before being fed into the 
    :func:`~pystiche_papers.johnson_alahi_li_2016.perceptual_loss`. Since this step is 
    necessary to get meaningful encodings from the 
    :func:`~pystiche_papers.johnson_alahi_li_2016.multi_layer_encoder`, the 
    pre-processing transform has to be learned within the output layer of the decoder. 
    To make this possible, ``150 * tanh(input)`` is used as activation in contrast to 
    the ``(tanh(input) + 1) / 2`` given in the paper.
```

docs/source/api/pystiche_papers.johnson_alahi_li_2016.rst
jbueltemeier: Yes, some of the top-level functions are not affected:

pystiche_papers.johnson_alahi_li_2016.preprocessor
pystiche_papers.johnson_alahi_li_2016.postprocessor
pystiche_papers.johnson_alahi_li_2016.image_loader
pystiche_papers.johnson_alahi_li_2016.images
pystiche_papers.johnson_alahi_li_2016.dataset

pystiche_papers/johnson_alahi_li_2016/_loss.py
jbueltemeier: Yeah, I think you're right about that, too. Especially because of the list of all functions that are affected by the parameter, it should be clear that `impl_params` is passed down if nothing else is described.

PR #109, opened 2020-08-03
Add paper impl_params permalinks

jbueltemeier:
Adding permalinks to the respective reference implementation parameters from the respective papers. Currently the permalinks are only added as a comment, do you have an idea how they can be better integrated? This is especially for those papers where the combination of `instance_norm` and `impl_params` is important.

For Example:
```python
"""
instance_norm:
https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/models/pyramid.lua#L3
not instance_norm:
https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/models/pyramid.lua#L3
"""
```

In Addition:
The permalinks of the `impl_params` setting of the pre-trained styles from the paper johnson_alahi_li_2016 are still missing

PR #121, opened 2020-08-04
change gatys_ecker_bethge from 2015 to 2016

### LINKED ISSUE #81: Change year of gatys_ecker_bethge from 2015 to 2016
While the [preprint was published in 2015](https://arxiv.org/abs/1508.06576) the [peer reviewed version was only published in 2016](https://ieeexplore.ieee.org/document/7780634). I think we should go with the latter and thus change the year in the namespace.

pmeier:
Closes #81


pmeier:
Blocked by #109

Issue #130, opened 2020-08-05
Move paper-independent modules to pystiche_papers.utils.modules


pmeier:
A few come to mind here:

- https://github.com/pmeier/pystiche_papers/blob/84fa874ca189761b8b54ef156a7e87ce7e98de21/pystiche_papers/johnson_alahi_li_2016/_data.py#L46

  https://github.com/pmeier/pystiche_papers/blob/84fa874ca189761b8b54ef156a7e87ce7e98de21/pystiche_papers/ulyanov_et_al_2016/_data.py#L32

- https://github.com/pmeier/pystiche_papers/blob/84fa874ca189761b8b54ef156a7e87ce7e98de21/pystiche_papers/ulyanov_et_al_2016/_modules.py#L14

- https://github.com/pmeier/pystiche_papers/blob/84fa874ca189761b8b54ef156a7e87ce7e98de21/pystiche_papers/ulyanov_et_al_2016/_modules.py#L66

pmeier:
Closed in #134

PR #134, opened 2020-08-05
Generalize paper independent modules 

jbueltemeier:
Move paper-independent modules as mentioned in  [#130](https://github.com/pmeier/pystiche_papers/issues/130).

### DIFF
--- a/pystiche_papers/johnson_alahi_li_2016/_data.py
+++ b/pystiche_papers/johnson_alahi_li_2016/_data.py
@@ -1,8 +1,11 @@
-from typing import Optional, Sized
+from typing import Optional, Sized, Tuple
 from urllib.parse import urljoin
 
+import torch
 from torch.utils.data import DataLoader, Dataset, Sampler
 
+import pystiche.image.transforms.functional as F
+from pystiche import image
 from pystiche.data import (
     DownloadableImage,
     DownloadableImageCollection,
@@ -11,7 +14,7 @@ from pystiche.data import (
 from pystiche.image import transforms
 
 from ..data.utils import FiniteCycleBatchSampler
-from ..utils.transforms import OptionalGrayscaleToFakegrayscale, TopLeftCropToMultiple
+from ..utils.transforms import OptionalGrayscaleToFakegrayscale
 
 __all__ = [
     "content_transform",
@@ -26,6 +29,20 @@ __all__ = [
 def content_transform(
     edge_size: int = 256, multiple: int = 16, impl_params: bool = True,
 ) -> transforms.ComposedTransform:
+    class TopLeftCropToMultiple(transforms.Transform):
+        def __init__(self, multiple: int):
+            super().__init__()
+            self.multiple = multiple
+
+        def calculate_size(self, input_image: torch.Tensor) -> Tuple[int, int]:
+            old_height, old_width = image.extract_image_size(input_image)
+            new_height = old_height - old_height % self.multiple
+            new_width = old_width - old_width % self.multiple
+            return new_height, new_width
+
+        def forward(self, input_image: torch.Tensor) -> torch.Tensor:
+            size = self.calculate_size(input_image)
+            return F.top_left_crop(input_image, size)
 
     transforms_ = [
         TopLeftCropToMultiple(multiple),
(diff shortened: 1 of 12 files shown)


pystiche_papers/johnson_alahi_li_2016/_data.py
```diff
@@ -28,28 +26,6 @@
 def content_transform(
     edge_size: int = 256, multiple: int = 16, impl_params: bool = True,
 ) -> transforms.ComposedTransform:
-    class TopLeftCropToMultiple(transforms.Transform):
```
pmeier, line 31: This was too ambitious. I don't think another paper will use this transform again. If time proves me wrong, we handle it then.

pystiche_papers/li_wand_2016/_data.py
```diff
@@ -18,11 +18,6 @@ def make_image_transform(image: str) -> transforms.ComposedTransform:
     if image == "emma":
         image_transform = transforms.Crop(origin=(30, 12), size=(930, 682))
     elif image == "jenny":
-
-        class MirrorHorizontally(transforms.Transform):
```
pmeier, line 22: Same as above. While this is not as specific, I don't think another paper will use it.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -5,50 +5,16 @@
...
-        self.out_channels = cast(Dict[str, nn.Module], self._modules)[
-            out_channel_name
-        ].out_channels
-
-
-def join_channelwise(*inputs: torch.Tensor, channel_dim: int = 1) -> torch.Tensor:
-    return torch.cat(inputs, dim=channel_dim)
-
-
-class AddNoiseChannels(nn.Module):
```
pmeier, line 33: Same as above.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -63,16 +29,6 @@ def upsample() -> nn.Upsample:
     return nn.Upsample(scale_factor=2.0, mode="nearest")
 
 
-class HourGlassBlock(SequentialWithOutChannels):
```
pmeier, line 66: On a second thought, we might just leave this here. My initial thought was that this is a commonly used structure, but with the current implementation I think its a bit too generic. Right now its basically an `nn.Sequential` with the three components. Thoughts?

pystiche_papers/utils/misc.py
```diff
@@ -25,6 +25,7 @@
     "same_size_output_padding",
     "is_valid_padding",
     "batch_up_image",
+    "join_channelwise",
```
pmeier, line 28: If we leave the `AddNoiseChannels` in the paper implementation, this should stay there as well.

tests/unit/utils/test_modules.py
```diff
@@ -34,3 +36,22 @@ def test_ResidualBlock_shortcut(double_module):
         double_module, shortcut=utils.ResidualBlock(double_module)
     )
     assert model(input) == ptu.approx(5.0)
+
+
+def test_SequentialWithOutChannels(subtests):
+    sequentialpaper = (nn.Conv2d(3, 3, 1), nn.Conv2d(3, 5, 1))
```
pmeier, line 42: Could you rename this to sequential_modules? This was accidentally introduced in b02df27b while refactoring.

tests/unit/utils/test_modules.py
```diff
@@ -34,3 +36,22 @@ def test_ResidualBlock_shortcut(double_module):
         double_module, shortcut=utils.ResidualBlock(double_module)
     )
     assert model(input) == ptu.approx(5.0)
+
+
+def test_SequentialWithOutChannels(subtests):
```
pmeier, line 41: Could you add another test checking whether `SequentialWithOutChannels` behaves just like an `nn.Sequential`?

tests/unit/utils/test_transforms.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_OptionalGrayscaleToFakegrayscale_no_grayscale(content_image):
+    transform = utils.OptionalGrayscaleToFakegrayscale()
+    actual = transform(content_image)
+    ptu.assert_allclose(actual, content_image)
+
+
+def test_OptionalGrayscaleToFakegrayscale():
+    image = torch.rand((1, 1, 32, 32))
```
pmeier, line 15: Please use the `content_image` together with `F.rgb_to_grayscale`-

PR #142, opened 2020-08-06
add replication tests for ulyanov_et_al_2016

pmeier:
Cc @jbueltemeier


pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -86,7 +86,7 @@ def images() -> DownloadableImageCollection:
             md5="dc9ad203263f34352e18bc29b03e1066",
             file="tuebingen_neckarfront__andreas_praefcke.jpg",
         ),
-        "CheHigh": DownloadableImage(
```
pmeier, line 89: I've aligned these keys with the style of the others.

PR #147, opened 2020-08-07
Add docs for li_wand_2016

jbueltemeier:


pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -48,6 +48,25 @@ def content_loss(
...
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            omitted, the default
+            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used. Defaults
+            to ``None``.
+        layer: Layer from which the encodings of the ``multi_layer_encoder`` should be
+            taken. Defaults to "relu4_2".
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to ``impl_params``. For details see below.
+
+    If ``impl_params is True`` , a score_weight of 2e1 is used instead of 1e0, and a
```
pmeier, line 66: Couldn't we move the first part into the description of `score_weight`? Maybe we should reserve the part below the `Args` for stuff that can't be explained within.

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+    r"""Style_loss from :cite:`LW2016`.
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            omitted,  the default
+            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used. Defaults
+            to ``None``.
```
pmeier, line 136:
```suggestion
            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used. 
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+        layers: Layers from which the encodings of the ``multi_layer_encoder`` should be
+            taken. If omitted, the defaults is used. Defaults to
+            ``("relu3_1", "relu4_1")``.
+        layer_weights: Layer weights of the operator. Defaults to ``"sum"``.
+        patch_size: Size of the patch.
```
pmeier, line 141: Please explain what happens if this is given as `int` rather than as tuple of two `int`s. See for example https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html?highlight=conv2d#torch.nn.Conv2d

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+        stride: Stride of the convolution. If omitted, the stride is determined with
```
pmeier, line 142: Same as above

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+            respect to `impl_params``. For details see below.
+        target_transforms: Optional augemntation transformations for the target. If
+            omitted, the transfomrms are determined with respect to `impl_params``. For
```
pmeier, line 145:
```suggestion
            omitted, the transforms are determined with respect to ``impl_params``. For
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+            details see below.
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to `impl_params``. For details see below.
+
+    If ``impl_params is True`` , a stride of 2 is used instead of 1, a score_weight of
+    1e-4 instead of the 1e0, and an additional score correction factor of 1.0 / 2.0 is
+    used instead of none. Additionally, no transformations are used in the scaling and
```
pmeier, line 152: The transforms are used regardless of `impl_params`. Only their parameters change.

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,36 @@ def style_loss(
...
+    rotation. In the paper two additional transformations are used in the scaling and
+    the rotation. Furthermore, normalized patches are used instead of the unnormalized
```
pmeier, line 154: You need to link to `extract_normalized_patches` or no one will understand what "normalized patches" are.

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -177,6 +239,21 @@ def perceptual_loss(
...
+    r"""Perceptual loss from :cite:`LW2016`.
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            omitted, the default
+            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used. Defaults
+            to ``None``.
```
pmeier, line 251:
```suggestion
            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used.
```

pystiche_papers/li_wand_2016/_nst.py
```diff
@@ -26,6 +26,30 @@ def nst(
...
+
+    Args:
+        content_image: Content image for the NST.
+        style_image: Style image for the NST.
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
+        criterion: Optimization criterion. If omitted, the default
+            :func:`~pystiche_papers.li_wand_2016.perceptual_loss` is used. Defaults to
+            ``None``.
```
pmeier, line 39: Not going to mark this again. Please apply this everywhere.

```suggestion
        criterion: Optimization criterion. If omitted, the default
            :func:`~pystiche_papers.li_wand_2016.perceptual_loss` is used.
```

pystiche_papers/li_wand_2016/_pyramid.py
```diff
@@ -14,6 +14,32 @@ def image_pyramid(
...
+            the paper. For details see below.
+        max_edge_size: Maximum edge size. Defaults to 384.
+        num_steps: Number of steps for each level. If omitted, the number is
+            determined with respect to ``impl_params``. For details see below. Defaults
+            to ``None``.
+        num_levels: Optional number of levels. If omitted, the number is determined by
+            the number of steps of factor two between ``max_edge_size`` and
+            ``min_edge_size`` or with respect to ``impl_params``. For details see below.
+        min_edge_size: Minimum edge size for the automatic calculation of
+            ``num_levels``. Defaults to 64.
```
pmeier, line 31: Please also wrap numbers in the literal blocks.

```suggestion
            ``num_levels``. Defaults to ``64``.
```

pystiche_papers/li_wand_2016/_pyramid.py
```diff
@@ -14,6 +14,32 @@ def image_pyramid(
     edge: Union[str, Sequence[str]] = "long",
     **octave_image_pyramid_kwargs: Any,
 ) -> pyramid.OctaveImagePyramid:
+    r"""Image pyramid from :cite:`LW2016`.
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        max_edge_size: Maximum edge size. Defaults to 384.
```
pmeier, line 23:
```suggestion
        max_edge_size: Maximum edge size. Defaults to ``384``.
```

pystiche_papers/li_wand_2016/_loss.py
jbueltemeier: Yes, but with `impl_param is True` the number of transformations is set to 1, so no additional transformation takes place. Or am I misunderstanding this right now?

pystiche_papers/li_wand_2016/_loss.py
jbueltemeier: Isn't it better to describe the influence of the parameter `impl_params` on the parameters uniformly in the block below `Args`?

pystiche_papers/li_wand_2016/_loss.py
pmeier: Where do you get this from?

pystiche_papers/li_wand_2016/_loss.py
pmeier: I think `Args` is the right place to describe everything related to the input arguments. For example, it bothers me that I don't know the default value for `score_weight` without reading on after the `Args` section.

pystiche_papers/li_wand_2016/_loss.py
jbueltemeier: From the original implementation. See [here](https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/cnnmrf.lua#L52) for the number of num_scales. For this reason the `num_scale_steps=1` if` impl_params is True` in this replication.

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -132,26 +132,36 @@ def style_loss(
...
         multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
             omitted,  the default
-            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used. 
+            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used.
         layers: Layers from which the encodings of the ``multi_layer_encoder`` should be
             taken. If omitted, the defaults is used. Defaults to
             ``("relu3_1", "relu4_1")``.
         layer_weights: Layer weights of the operator. Defaults to ``"sum"``.
-        patch_size: Size of the patch.
+        patch_size: Size of the patch. Defaults to ``3``.
```
pmeier, line 140:
```suggestion
        patch_size: Size of a patch. Defaults to ``3``.
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -223,7 +233,7 @@ def regularization(
             results. Defaults to ``2.0``.
         score_weight: Score weight of the operator. Defaults to ``1e-3``.
 
-    If ``impl_params is True`` , an additional score correction factor of 1.0 / 2.0
+    If ``impl_params is True`` , an additional score correction factor of ``1.0 / 2.0``
     is used instead of none.
```
pmeier, line 237:
```suggestion
    is used.
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -132,26 +132,36 @@ def style_loss(
...
+    is used instead of none. Additionally, no transformations are used in the scaling
+    and rotation. In the paper two additional transformations are used in the scaling
+    and the rotation. Furthermore, normalized patches are used instead of the
+    unnormalized  patches. For details see
+    `extract_normalized_patches2d <https://github.com/pmeier/pystiche_papers/blob/b15202068d683c8b60f8a437e562f3d922fe0f3e/pystiche_papers/li_wand_2016/_utils.py#L51>`_
+    .
+
+    The parameters ``patch_size`` and ``stride`` can either be:
+
+    * a single int – in which case the same value is used for the height and width
```
pmeier, line 161:
```suggestion
    * a single :class:`int` – in which case the same value is used for the height and width
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -132,26 +132,36 @@ def style_loss(
...
-    1e-4 instead of the 1e0, and an additional score correction factor of 1.0 / 2.0 is
-    used instead of none. Additionally, no transformations are used in the scaling and
-    rotation. In the paper two additional transformations are used in the scaling and
-    the rotation. Furthermore, normalized patches are used instead of the unnormalized
-    patches.
+            determined with respect to `impl_params``. Defaults to ``1e-4`` if
+            ``impl_params is True`` otherwise ``1e0``.
+
+    If ``impl_params is True`` , an additional score correction factor of ``1.0 / 2.0``
+    is used instead of none. Additionally, no transformations are used in the scaling
```
pmeier, line 152:
```suggestion
    is used instead. Additionally, no transformations are used in the scaling
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,45 @@ def style_loss(
...
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to `impl_params``. Defaults to ``1e-4`` if
+            ``impl_params is True`` otherwise ``1e0``.
+
+    If ``impl_params is True`` , an additional score correction factor of ``1.0 / 2.0``
+    is used instead of none. Additionally, no transformations are used in the scaling
+    and rotation. In the paper two additional transformations are used in the scaling
+    and the rotation. Furthermore, normalized patches are used instead of the
+    unnormalized  patches. For details see
+    `extract_normalized_patches2d <https://github.com/pmeier/pystiche_papers/blob/b15202068d683c8b60f8a437e562f3d922fe0f3e/pystiche_papers/li_wand_2016/_utils.py#L51>`_
```
pmeier, line 156: Please use a `:func:` link

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,45 @@ def style_loss(
...
+    .
+
+    The parameters ``patch_size`` and ``stride`` can either be:
+
+    * a single int – in which case the same value is used for the height and width
+      dimension
+    * a tuple of two ints – in which case, the first int is used for the height
+      dimension, and the second int for the width dimension
```
pmeier, line 164:
```suggestion
    * a tuple of two ints – in which case, the first int is used for the vertical
      dimension, and the second int for the horizontal dimension
```

pystiche_papers/li_wand_2016/_loss.py
pmeier: With `num_scale_steps == 1` you will have three transformations. Nonetheless, the reference implementation seems to default to `0` steps. I got the parameters from 

https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L65-L70

since this result is included in the paper. Thoughts?

jbueltemeier: I think we should change this to `0`, as the author explicitly specifies the default value as `0`.
See  [here](https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L17).

pmeier: I think this OK for now, but we should keep this in mind for the replication. Could you send a fix?

jbueltemeier: Just to be on the safe side, do you mean a fix by setting the default value to `0` and the description here in the doc is ok for now?

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,44 @@ def style_loss(
...
+            the paper. For details see below.
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            omitted,  the default
+            :func:`~pystiche_papers.li_wand_2016.multi_layer_encoder` is used.
+        layers: Layers from which the encodings of the ``multi_layer_encoder`` should be
+            taken. If omitted, the defaults is used. Defaults to
+            ``("relu3_1", "relu4_1")``.
+        layer_weights: Layer weights of the operator. Defaults to ``"sum"``.
+        patch_size: Size of a patch. Defaults to ``3``.
+        stride: Stride of the convolution. If omitted, the stride is determined with
```
pmeier, line 141:
```suggestion
        patch_size: Size of the sliding window. Defaults to ``3``.
        stride: Stride of the sliding window. If omitted, the stride is determined with
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,44 @@ def style_loss(
...
+    and rotation. In the paper two additional transformations are used in the scaling
+    and the rotation. Furthermore, normalized patches are used instead of the
+    unnormalized  patches. For details see
+    :func:`~pystiche_papers.li_wand_2016._utils.extract_normalized_patches2d`.
+
+    The parameters ``patch_size`` and ``stride`` can either be:
+
+    * a single :class:`int` – in which case the same value is used for the height and
+      width dimension
+    * a tuple of two ints – in which case, the first int is used for the vertical
```
pmeier, line 162:
```suggestion
    * a tuple of two :class:`int` s – in which case, the first int is used for the vertical
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,44 @@ def style_loss(
...
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to `impl_params``. Defaults to ``1e-4`` if
+            ``impl_params is True`` otherwise ``1e0``.
+
+    If ``impl_params is True`` , an additional score correction factor of ``1.0 / 2.0``
+    is used. Additionally, no transformations are used in the scaling
+    and rotation. In the paper two additional transformations are used in the scaling
+    and the rotation. Furthermore, normalized patches are used instead of the
+    unnormalized  patches. For details see
+    :func:`~pystiche_papers.li_wand_2016._utils.extract_normalized_patches2d`.
```
pmeier, line 156: We should refactor this is list to make it more legible

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -105,6 +124,45 @@ def style_loss(
...
+            determined with respect to `impl_params``. Defaults to ``1e-4`` if
+            ``impl_params is True`` otherwise ``1e0``.
+
+    If ``impl_params is True``:
+
+    * an additional score correction factor of ``1.0 / 2.0`` is used
+    * Normalized patches are used instead of the unnormalized patches. For details see
+      :func:`~pystiche_papers.li_wand_2016._utils.extract_normalized_patches2d`.
+    * No transformations are used instead of two additional transformations in the
+      scaling and the rotation.
```
pmeier, line 157:
```suggestion
    If ``impl_params is True``,

    * an additional score correction factor of ``1.0 / 2.0`` is used,
    * normalized patches are used (see :func:`~pystiche_papers.li_wand_2016._utils.extract_normalized_patches2d` for details), and
    * no target augmentation transformations are used.
```

pystiche_papers/li_wand_2016/_loss.py
```diff
@@ -165,6 +223,19 @@ def calculate_score(self, input_repr: torch.Tensor) -> torch.Tensor:
 def regularization(
     impl_params: bool = True, exponent: float = 2.0, score_weight: float = 1e-3,
 ) -> TotalVariationOperator:
+    r"""Regularization from :cite:`LW2016`.
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see below.
+        exponent: Parameter :math:`\beta` . A higher value leads to more smoothed
```
pmeier, line 232: Could you check if `\beta` is even used in the paper? I don't recall that.

jbueltemeier: `\beta` is not used but nothing else is used either.
> For this reason, we penalize the squared gradient norm [20] to encourage smoothness in the synthesized image:

pmeier: In that case I would simply remove the first sentence.

pmeier:
Please do not merge this yet. Something is going on with the documentation build in the CI. I'll look into it.

pystiche_papers/li_wand_2016/_loss.py
pmeier: Correct. For the replication though, we should use `num_scale_steps = num_rotation_steps = 1` to get closer to the result in the original replication.

PR #144, opened 2020-08-07
Add doc for ulyanov_et_al_2016

jbueltemeier:

PR #157, opened 2020-08-14
Add docs for ulyanov_et_al_2016 general

jbueltemeier:


pmeier:
Partially supersedes #144

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,95 @@
...
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.transformer`.
+
+
+.. _table-branches-ulyanov_et_al_2016:
+
+The original authors have provided two implementations. Unfortunately, the
+hyperparameters used differ from those in the paper, as well as
+among the individual implementations themselves. If you use ``instance_norm`` and
+``imp_params`` as in the following table, the appropriate parameters will be used:
```
pmeier, line 26:
```suggestion
``impl_params`` as in the following table, the appropriate parameters will be used:
```

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,95 @@
...
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.style_transform`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.batch_sampler`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.content_loss`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.transformer`.
+
+
+.. _table-branches-ulyanov_et_al_2016:
+
+The original authors have provided two implementations. Unfortunately, the
```
pmeier, line 23: IIRC the authors published two papers: one without and a later one with instance normalization. If that is the case, we should describe this properly. Although they live in the same repo, each branch is an independent reference implementation for a paper.

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -50,6 +50,29 @@ def content_loss(
...
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see
+            :ref:`here <table-hyperparameters-ulyanov_et_al_2016>` and below.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            :ref:`here <table-branches-ulyanov_et_al_2016>`.
```
pmeier, line 63: Basically we are switching between two reference implementations, correct? If that is the case we should write it like that since the fact that they live on two git branches is irrelevant.

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -50,6 +50,29 @@ def content_loss(
...
+        multi_layer_encoder: Pretrained :class:`~pystiche.enc.MultiLayerEncoder`. If
+            omitted, the default
+            :func:`~pystiche_papers.ulyanov_et_al_2016.multi_layer_encoder` is used.
+        layer: Layer from which the encodings of the ``multi_layer_encoder`` should be
+            taken. Defaults to ``"relu4_2"``.
+        score_weight: Score weight of the operator. If omitted, the ``score_weight`` is
+            determined with respect to ``impl_params`` and ``instance_norm``. For
+            details see :ref:`here <table-hyperparameters-ulyanov_et_al_2016>`.
+
+    If ``impl_params is True`` , the score is divided once more by the batch_size.
```
pmeier, line 73:
```suggestion
    If ``impl_params is True`` , the score is divided twice by the batch_size.
```

pystiche_papers/ulyanov_et_al_2016/_utils.py
```diff
@@ -34,19 +35,43 @@ def postprocessor() -> transforms.CaffePostprocessing:
...
+             :class:`~torch.optim.lr_scheduler.ExponentialLR`.
+    """
     last_epoch: int
     gamma: float
     base_lrs: List[float]
 
     def __init__(
         self, optimizer: Optimizer, gamma: float, delay: int, **kwargs: Any
     ) -> None:
+
```
pmeier, line 74:
```suggestion
```

pystiche_papers/ulyanov_et_al_2016/_utils.py
```diff
@@ -59,6 +84,15 @@ def get_lr(self) -> List[float]:  # type: ignore[override]
...
 
 def lr_scheduler(optimizer: Optimizer, impl_params: bool = True,) -> ExponentialLR:
+    r"""Learning rate scheduler from :cite:`ULVL2016`.
+
+    Args:
+        optimizer: Wrapped optimizer.
+        impl_params: If ``True``, an :class:`~torch.optim.lr_scheduler.ExponentialLR`
+        with a ``gamma of 0.8`` is used instead of a
+        :func:`~pystiche_papers.ulyanov_et_al_2016.DelayedExponentialLR` with a
+        ``gamma of 0.7`` and a ``delay of 5``.
```
pmeier, line 94:
```suggestion
        with ``gamma==0.8`` is used instead of a
        :func:`~pystiche_papers.ulyanov_et_al_2016.DelayedExponentialLR` with
        ``gamma==0.7`` and ``delay==5``.
```

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,103 @@
...
+
+  - the :func:`~pystiche_papers.ulyanov_et_al_2016.content_loss`,
+  - the :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`, as well as
+
+- the ``layers`` for
+
+  - the :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`.
+
+The ``layers`` have either ``4 layers`` ``("relu1_1", "relu2_1", "relu3_1", "relu4_1")``
+or ``5 Layer`` ``("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1")``.
```
pmeier, line 75: I see what you are doing, but I'm not happy with `4 layers` etc key. Maybe call it configuration `1` and `2`?

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,103 @@
...
+
++-----------------+-----------+------------------------------------+
+|                 |           |        ``instance_norm``           |
++-----------------+-----------+-------------+----------------------+
+|                 |           |   ``True``  |       ``False``      |
++-----------------+-----------+-------------+----------------------+
+| ``impl_params`` | ``True``  | ``master``  | ``texture_nets_v1``  |
+|                 +-----------+-------------+----------------------+
+|                 | ``False`` | ``"paper"`` |      ``"paper"``     |
++-----------------+-----------+-------------+----------------------+
```
pmeier, line 51: I think this table is obsolete. It does not add any information to the explanation above.

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,103 @@
...
+among the individual implementations themselves. If you use ``instance_norm`` and
+``impl_params`` as in the following table, the appropriate parameters will be used:
+
+  - ``master``: The parameters specified in the implementation branch
+    `master <https://github.com/pmeier/texture_nets/tree/master>`_.
+  - ``texture_nets_v1``: The parameters specified in the implementation branch
+    `texture_nets_v1 <https://github.com/pmeier/texture_nets/tree/texture_nets_v1>`_.
+  - ``"paper"``: The parameters specified in the papers. Here ``instance_norm`` is only
+    used to use either :class:`~torch.nn.InstanceNorm2d` or
+    :class:`~torch.nn.BatchNorm2d`.
```
pmeier, line 41: I would place this as a regular sentence (not a list item) below.

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,103 @@
...
+- the ``layers`` for
+
+  - the :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`.
+
+The ``layers`` have either ``4 layers`` ``("relu1_1", "relu2_1", "relu3_1", "relu4_1")``
+or ``5 Layer`` ``("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1")``.
+
+
++-------------------+---------------+--------------+-----------------------+
+|   Parameter       |  ``"paper"``  |  ``master``  |  ``texture_nets_v1``  |
```
pmeier, line 79: Do we need the `paper` column? I think we should only list the reference implementation. At least that is what we did with [`johnson_alahi_li_2016`](https://github.com/pmeier/pystiche_papers/blob/master/docs/source/api/pystiche_papers.johnson_alahi_li_2016.rst).

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,103 @@
...
++===================+===============+==============+=======================+
+|  ``num_epochs``   |      10       |      10      |           25          |
++-------------------+---------------+--------------+-----------------------+
+|     ``lr``        |     1e-1      |    1e-3      |         1e-1          |
++-------------------+---------------+--------------+-----------------------+
+|  ``num_batches``  |     200       |     2000     |          300          |
++-------------------+---------------+--------------+-----------------------+
+|  ``batch_size``   |      16       |       1      |           4           |
++-------------------+---------------+--------------+-----------------------+
+| ``content_loss``  |               |              |                       |
```
pmeier, line 89: I don't think we should have sections within this table. Maybe `content_score_weight`? Same for the `style_loss` below.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -39,6 +39,42 @@ def training(
         Callable[[int, Union[torch.Tensor, pystiche.LossDict], float, float], None]
     ] = None,
 ) -> nn.Module:
+    r"""Training a transformer for the NST.
+
+    Args:
+        content_image_loader: Content images used as input for the ``transformer``.
+        style: Style image on which the ``transformer`` should be trained. If the
+            input is an string, the style image is read from the images in
+            :func:`~pystiche_papers.ulyanov_et_al_2016.images`.
```
pmeier, line 48:
```suggestion
            input is an string, the style image is read from
            :func:`~pystiche_papers.ulyanov_et_al_2016.images`.
```

pmeier (changes requested):

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
jbueltemeier: I think if the column `"paper"` is removed, this sentence is completely superfluous, since `instance_norm` is already described in the documentation. Or what do you think?

pmeier: Yeah, you could remove this here. In that case I would make the sentence above L23-24 more clear, i.e. completely the same parameters and only batch norm is switched for instance norm.

jbueltemeier: I just found the Supplementary material (see [Supplementary material](http://sites.skoltech.ru/app/data/uploads/sites/25/2017/01/texture_nets_v2_sup.pdf)) and I think that this is not true as described now. The `texture_net_v2` is not the same architecture as in the first paper. However, in the second paper the authors experimented with the structure and replaced the `BatchNorm2d `with `InstanceNorm2d`. (See [paper](https://arxiv.org/pdf/1701.02096.pdf) section 5.1). I will rephrase it like this.

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,90 @@
...
+
+
++---------------------------+--------------+-----------------------+
+|   Parameter               |  ``master``  |  ``texture_nets_v1``  |
++===========================+==============+=======================+
+|  ``num_epochs``           |      10      |           25          |
++---------------------------+--------------+-----------------------+
+|     ``lr``                |    1e-3      |         1e-1          |
++---------------------------+--------------+-----------------------+
+|  ``num_batches``          |     2000     |          300          |
```
pmeier, line 76: The indentation is off. Please check the complete table
```suggestion
| ``num_batches``           |     2000     |          300          |
```

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,90 @@
...
+``("relu1_1", "relu2_1", "relu3_1", "relu4_1")`` or  configuration ``2``
+``("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1")``.
+
+
++---------------------------+--------------+-----------------------+
+|   Parameter               |  ``master``  |  ``texture_nets_v1``  |
++===========================+==============+=======================+
+|  ``num_epochs``           |      10      |           25          |
++---------------------------+--------------+-----------------------+
+|     ``lr``                |    1e-3      |         1e-1          |
```
pmeier, line 74:
```suggestion
| ``learning_rate``                |    1e-3      |         1e-1          |
```

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,90 @@
...
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`,
+  - :func:`~pystiche_papers.ulyanov_et_al_2016.transformer`.
+
+
+.. _table-branches-ulyanov_et_al_2016:
+
+The original authors experimented with the network architecture and described two
+versions of the network in different papers, replacing the
+:class:`~torch.nn.BatchNorm2d` by :class:`~torch.nn.InstanceNorm2d` as an improvement.
+They call it StyleNet with a postfix BN if it is equipped with
```
pmeier, line 26: "equipped" seems weird here. Could you rephrase?

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -39,6 +39,42 @@ def training(
         Callable[[int, Union[torch.Tensor, pystiche.LossDict], float, float], None]
     ] = None,
 ) -> nn.Module:
+    r"""Training a transformer for the NST.
+
+    Args:
+        content_image_loader: Content images used as input for the ``transformer``.
+        style: Style image on which the ``transformer`` should be trained. If the
+            input is an string, the style image is read from
```
pmeier, line 47:
```suggestion
            input is :class:`str`, the style image is read from
```

pmeier (approved):
Only nitpicks

pmeier:
Failing tests are unrelated. I'll fix this in a moment.

PR #158, opened 2020-08-14
Add docs for ulyanov_et_al_2016 modules

jbueltemeier:
Partially supersedes [#144](https://github.com/pmeier/pystiche_papers/pull/144).

### DIFF
--- a/docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
+++ b/docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
@@ -61,7 +61,7 @@ The following table provides an overview of the parameters:
 
   - the :func:`~pystiche_papers.ulyanov_et_al_2016.style_loss`.
 
-The ``layers`` have either configuration ``1``
+The ``style_layers`` have either configuration ``1``
 ``("relu1_1", "relu2_1", "relu3_1", "relu4_1")`` or  configuration ``2``
 ``("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1")``.
 
@@ -83,6 +83,7 @@ The ``layers`` have either configuration ``1``
 +---------------------------+-------------+---------------------+
 | ``style_layers``          | ``1``       | ``2``               |
 +---------------------------+-------------+---------------------+
+
 
 .. automodule:: pystiche_papers.ulyanov_et_al_2016
 
--- a/docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
+++ b/docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
@@ -84,7 +84,6 @@ The ``layers`` have either configuration ``1``
 | ``style_layers``          | ``1``       | ``2``               |
 +---------------------------+-------------+---------------------+
 
-
 .. automodule:: pystiche_papers.ulyanov_et_al_2016
 
 .. autofunction:: batch_sampler
--- a/pystiche_papers/ulyanov_et_al_2016/_modules.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_modules.py
@@ -16,6 +16,13 @@ def join_channelwise(*inputs: torch.Tensor, channel_dim: int = 1) -> torch.Tenso
 
 
 class AddNoiseChannels(nn.Module):
+    r"""Adds noise channels in the size of the input to the input.
+
+    Args:
+        in_channels: Number of input channels.
+        num_noise_channels: Number of additional noise channels. Defaults to 3.
+    """
+
     def __init__(
         self, in_channels: int, num_noise_channels: int = 3,
     ):
@@ -49,6 +56,12 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    Args:
+        intermediate: Middle :class:`~torch.nn.Module` of the block.
+    """
+
     def __init__(self, intermediate: nn.Module):
         modules = (
             ("down", downsample()),
@@ -86,6 +99,24 @@ def activation(
 
 
 class ConvBlock(SequentialWithOutChannels):
+    r"""ConvBlock from :cite:`ULVL2016`.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels:  Number of channels produced by the convolution.
+        kernel_size: Size of the convolving kernel.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        stride: Stride of the convolution. Defaults to 1.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
+
+    """
+
     def __init__(
         self,
         in_channels: int,
@@ -120,6 +151,22 @@ class ConvBlock(SequentialWithOutChannels):
 
 
 class ConvSequence(SequentialWithOutChannels):
+    r"""Sequence of convolutional blocks that occurs repeatedly in :cite:`ULVL2016`.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels: Number of channels produced by the convolution
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
+
+    """
+
     def __init__(
         self,
         in_channels: int,
@@ -150,6 +197,20 @@ class ConvSequence(SequentialWithOutChannels):
 
 
 class JoinBlock(nn.Module):
+    r"""JoinBlock from :cite:`ULVL2016`.
+
+    Args:
+        branch_in_channels: Number of channels in the branch input.
+        names: Optional names for the blocks. If omitted, the blocks are numbered.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        channel_dim: The dimension over which the tensors are concatenated. Defaults to
+            1.
+
+    """
+
     def __init__(
         self,
         branch_in_channels: Sequence[int],
@@ -188,6 +249,17 @@ class JoinBlock(nn.Module):
 
 
 class BranchBlock(nn.Module):
+    r"""BranchBlock from :cite:`ULVL2016`.
+
+    Args:
+        deep_branch: Input from the branch one step deeper in the pyramid.
+        shallow_branch: Input from the current branch.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+    """
+
     def __init__(
         self,
         deep_branch: nn.Module,
@@ -224,6 +296,24 @@ def level(
     num_noise_channels: int = 3,
     inplace: bool = True,
 ) -> SequentialWithOutChannels:
+    r"""Defines one level of the Transformer from :cite:`ULVL2016`.
+
+    Args:
+        prev_level_block: Optional Input from the previous level. If ``None``, only one
+            ConvSequence is returned.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        in_channels: Number of channels in the input image. Defaults to 3.
+        num_noise_channels: Number of additional noise channels. Defaults to 3.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
+
+    """
+
     def conv_sequence(
         in_channels: int, out_channels: int, use_noise: bool = False
     ) -> SequentialWithOutChannels:
@@ -340,4 +430,18 @@ def transformer(
     instance_norm: bool = True,
     levels: int = 6,
 ) -> Transformer:
+    r"""Initialized the Transformer from :cite:`ULVL2016`.
+
+    Args:
+        style: FIXME this should be removed.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        levels: Number of levels in the Transformer. Defaults to 6.
+
+    """
     return Transformer(levels, impl_params=impl_params, instance_norm=instance_norm)
--- a/pystiche_papers/ulyanov_et_al_2016/_modules.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_modules.py
@@ -276,12 +276,12 @@ class JoinBlock(nn.Module):
 class BranchBlock(nn.Module):
     r"""BranchBlock from :cite:`ULVL2016`.
 
-    This block combines the ``deep_branch`` with the ``shallow_branch`` using a
+    Joins the branch from the previous pyramid with the current level using a
     :class:`~pystiche_paper.ulyanov_et_al_2016._modules.JoinBlock`.
 
     Args:
-        deep_branch: Input from the branch one step deeper in the pyramid.
-        shallow_branch: Input from the current branch.
+        deep_branch: Input from the previous pyramid.
+        shallow_branch: Input from the current level.
         instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
             :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
             flag is used for switching between two reference implementations. For
--- a/pystiche_papers/ulyanov_et_al_2016/_modules.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_modules.py
@@ -59,7 +59,8 @@ class HourGlassBlock(SequentialWithOutChannels):
     r"""HourGlassBlock from :cite:`ULVL2016`.
 
     Args:
-        intermediate: Middle :class:`~torch.nn.Module` of the block.
+        intermediate: Middle :class:`~torch.nn.Module` of the block. The
+        ``out_channels`` is taken from ``intermediate``.
     """
 
     def __init__(self, intermediate: nn.Module):
@@ -107,12 +108,13 @@ class ConvBlock(SequentialWithOutChannels):
         kernel_size: Size of the convolving kernel.
         impl_params: If ``True``, use the parameters used in the reference
             implementation of the original authors rather than what is described in
-            the paper. For details see FIXME.
+            the paper. For details see
+            :ref:`here <table-hyperparameters-ulyanov_et_al_2016>`.
         stride: Stride of the convolution. Defaults to ``1``.
         instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
             :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
-            flag is used for switching between the github branches. For details see
-            FIXME.
+            flag is used for switching between two reference implementations. For
+            details see :ref:`here <table-branches-ulyanov_et_al_2016>`.
         inplace: If ``True`` perform the activation in-place.
 
     The parameters ``kernel_size`` and ``stride`` can either be:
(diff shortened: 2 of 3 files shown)


pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -86,6 +99,31 @@ def activation(
...
 class ConvBlock(SequentialWithOutChannels):
+    r"""ConvBlock from :cite:`ULVL2016`.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels:  Number of channels produced by the convolution.
+        kernel_size: Size of the convolving kernel.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
```
pmeier, line 110: Address the `FIXME` here and everywhere else please.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -86,6 +99,31 @@ def activation(
...
+        stride: Stride of the convolution. Defaults to ``1``.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
```
pmeier, line 116:
```suggestion
        inplace: If ``True`` perform the activation in-place.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -120,6 +158,22 @@ def __init__(
 
 
 class ConvSequence(SequentialWithOutChannels):
+    r"""Sequence of convolutional blocks that occurs repeatedly in :cite:`ULVL2016`.
```
pmeier, line 161: Since this not mentioned in the paper, please explain what each sequence comprises.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -120,6 +158,22 @@ def __init__(
...
+        in_channels: Number of channels in the input.
+        out_channels: Number of channels produced by the convolution
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
```
pmeier, line 173: See above

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +204,20 @@ def conv_block(
 
 
 class JoinBlock(nn.Module):
+    r"""JoinBlock from :cite:`ULVL2016`.
```
pmeier, line 207: Just to be sure: could you check the paper and verify that the join blocks in the paper comprise the norm layers?

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -188,6 +256,17 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor:
 
 
 class BranchBlock(nn.Module):
+    r"""BranchBlock from :cite:`ULVL2016`.
```
pmeier, line 259: Same as with `ConvSequence`.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -224,6 +303,24 @@ def level(
     num_noise_channels: int = 3,
     inplace: bool = True,
 ) -> SequentialWithOutChannels:
+    r"""Defines one level of the Transformer from :cite:`ULVL2016`.
```
pmeier, line 306:
```suggestion
    r"""Defines one level of the transformer from :cite:`ULVL2016`.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -224,6 +303,24 @@ def level(
...
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        in_channels: Number of channels in the input image. Defaults to ``3``.
+        num_noise_channels: Number of additional noise channels. Defaults to ``3``.
+        inplace: Can optionally do the operation in-place. Defaults to ``True``.
```
pmeier, line 320: Same as above.

pystiche_papers/ulyanov_et_al_2016/_modules.py
pmeier, line 306: Same as with `ConvSequence`.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -340,4 +437,18 @@ def transformer(
...
+    Args:
+        style: FIXME this should be removed.
+        impl_params: If ``True``, use the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper. For details see FIXME.
+        instance_norm: If ``True``, use :class:`~torch.nn.InstanceNorm2d` rather than
+            :class:`~torch.nn.BatchNorm2d` as described in the paper. Additionally this
+            flag is used for switching between the github branches. For details see
+            FIXME.
+        levels: Number of levels in the Transformer. Defaults to ``6``.
```
pmeier, line 451:
```suggestion
        levels: Number of levels in the transformer. Defaults to ``6``.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -340,4 +437,18 @@ def transformer(
     instance_norm: bool = True,
     levels: int = 6,
 ) -> Transformer:
+    r"""Transformer from :cite:`ULVL2016`.
+
+    Args:
+        style: FIXME this should be removed.
```
pmeier, line 443: Correct.

```suggestion
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
jbueltemeier: The join blocks contain the norm layers, but also an upsample block. I will indicate this in the documentation that this upsample block is not included.

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -1,6 +1,10 @@
 ``pystiche_papers.ulyanov_et_al_2016``
 =========================================
 
+.. _table-branches-ulyanov_et_al_2016:
+
+.. _table-hyperparameters-ulyanov_et_al_2016:
+
```
pmeier, line 7: This should be resolved when #157 lands. Thus, you should remove this and simply rebase when its ready.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -120,6 +160,30 @@ def __init__(
...
 
 class ConvSequence(SequentialWithOutChannels):
+    r"""Sequence of convolutional blocks that occurs repeatedly in :cite:`ULVL2016`.
+
+    Each sequence contains three ``ConvBlock`` from :cite:`ULVL2016`:
+
+    * A convolution with ``in_channels`` as input and ``out_channels`` as output and a
+      ``kernel_size=3``.
+    * A convolution with ``out_channels`` as input and output and a ``kernel_size=3``.
+    * A convolution with ``out_channels`` as input and output and a ``kernel_size=1``.
```
pmeier, line 170: I think you can shorten this to say "three sequential conv blocks while the first two have a kernel size of 3 ...". Conv block should link to the class defined above.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -86,6 +100,32 @@ def activation(
 
 
 class ConvBlock(SequentialWithOutChannels):
+    r"""ConvBlock from :cite:`ULVL2016`.
```
pmeier, line 103: We should also include a short explanation here.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +214,24 @@ def conv_block(
 
 
 class JoinBlock(nn.Module):
+    r"""JoinBlock from :cite:`ULVL2016`.
+
+    This block concatenate two inputs along the ``channel _dim`` with previous
```
pmeier, line 219: 1. In cats an arbitrary number of inputs.
2. 
```suggestion
    This block concatenate two inputs along the ``channel _dim`` with prefixed
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +214,24 @@ def conv_block(
...
+    normalization modules. It therefore corresponds to the ``JoinBlock`` from the paper
+    only without upsample block.
+
+    Args:
+        branch_in_channels: Number of channels in the branch input.
```
pmeier, line 224: I wonder if we should rename this to `in_channels` since this block is not restricted to work with branches. Might be good idea for a follow-up PR.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +214,24 @@ def conv_block(
...
+        names: Optional names for the blocks. If omitted, the blocks are numbered.
```
pmeier, line 225:
```suggestion
        names: Optional names for the normalization modules. If omitted, the modules will be numbered.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -188,6 +270,20 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor:
 
 
 class BranchBlock(nn.Module):
+    r"""BranchBlock from :cite:`ULVL2016`.
+
+    This block combines the ``deep_branch`` with the ``shallow_branch`` using a
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.JoinBlock`.
```
pmeier, line 276: Isn't this basically a higher level, i. e. not the first, from the paper? If that is the case, I think its more clear to say `deep_branch` is the previous and `shallow_branch` is the current level.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -115,10 +115,7 @@ def stylization(
 ) -> torch.Tensor:
     device = input_image.device
     if isinstance(transformer, str):
-        style = transformer
-        transformer = _transformer(
-            style=style, impl_params=impl_params, instance_norm=instance_norm,
-        )
+        transformer = _transformer(impl_params=impl_params, instance_norm=instance_norm)
```
pmeier, line 118: I totally forgot about that. The parameter `style` should be available for the `transformer` since it will be used to load the pretrained weights. Could you please revert the change?

pmeier (changes requested):
Lets wait until #157 lands and continue to work on this afterwards

pystiche_papers/ulyanov_et_al_2016/_modules.py
jbueltemeier: Basically this block combines only these two branches with the help of the `JoinBlock`. Where the branches come from is not known to the block. That's why I'd leave it so general. That the `deep_branch `comes from a previous level is defined in `level`. The naming of the branches comes from the paper.

pmeier: Yeah, but this block has no other purpose. I can't imagine a use-case where it is used otherwise. Furthermore, the names `deep_branch` and `shallow_branch` hint at a very specific purpose. So we should either rename those or make the doc string more specific. I'm voting for the latter. Maybe something like "joins the branch from the previous level with the current"

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -16,6 +16,13 @@ def join_channelwise(*inputs: torch.Tensor, channel_dim: int = 1) -> torch.Tenso
 
 
 class AddNoiseChannels(nn.Module):
+    r"""Adds noise channels in the size of the input to the input.
```
pmeier, line 19:
```suggestion
    r"""Adds white noise channels to the input.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -49,6 +56,13 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    Args:
+        intermediate: Middle :class:`~torch.nn.Module` of the block. The
+        ``out_channels`` is taken from ``intermediate``.
```
pmeier, line 63:
```suggestion
    Args:
        intermediate: Module in between the down- and upsampling.
        
    Attributes:
        out_channels: ``ìntermediate.out_channels``
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -86,6 +100,38 @@ def activation(
 
 
 class ConvBlock(SequentialWithOutChannels):
+    r"""ConvBlock from :cite:`ULVL2016`.
+
+    This block contains a :class:`~torch.nn.ReflectionPad2d` followed by a
+    :class:`~torch.nn.Conv2d` with subsequent normalization depending on
+    ``instance_norm`` and an activation function depending on ``impl_params`` and
+    ``instance_norm``. The activation function is either a :class:`~torch.nn.ReLU` or a
+    :class:`~torch.nn.LeakyReLU` with ``slope=0.01``.
```
pmeier, line 109: This is a little too detailed. It basically needs to only state that:

- convolution is followed by normalization and activation
- image is reflection-padded to preserve the size

Furthermore, the type of activation should be put below the `Args` section, since it depends on `instance_norm`.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -120,6 +166,27 @@ def __init__(
 
 
 class ConvSequence(SequentialWithOutChannels):
+    r"""Sequence of convolutional blocks that occurs repeatedly in :cite:`ULVL2016`.
+
+    Each sequence contains three
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.ConvBlock` blocks while the
+    first two have a ``kernel_size of 3`` and the last one a ``kernel_size of 1``.
```
pmeier, line 173: The "``<text>``" are literal blocks commonly used to display code. "kernel_size of 3" is not valid python code. 

```suggestion
    first two uses ``kernel_size == 3``, the last one uses ``kernel_size == 1``.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +217,25 @@ def conv_block(
 
 
 class JoinBlock(nn.Module):
+    r"""JoinBlock from :cite:`ULVL2016`.
+
+    This block cats an arbitrary number of inputs along the ``channel _dim`` with
```
pmeier, line 222: I only used `cat` to shorten my comment :smile:

```suggestion
    This block concatenates an arbitrary number of inputs along the ``channel _dim`` with
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +217,25 @@ def conv_block(
...
+    prefixed normalization modules. It therefore corresponds to the ``JoinBlock`` from
+    the paper only without upsample block.
```
pmeier, line 224:
```suggestion
    prefixed normalization modules.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -150,6 +217,25 @@ def conv_block(
 
 
 class JoinBlock(nn.Module):
+    r"""JoinBlock from :cite:`ULVL2016`.
```
pmeier, line 220: I think this is a better fit here.

```suggestion
    r"""JoinBlock from :cite:`ULVL2016` without upsampling.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -188,6 +274,20 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor:
 
 
 class BranchBlock(nn.Module):
+    r"""BranchBlock from :cite:`ULVL2016`.
+
+    Joins the branch from the previous pyramid with the current level using a
```
pmeier, line 279:
```suggestion
    Joins the branch from the previous pyramid level with the current using a
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -188,6 +274,20 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor:
...
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.JoinBlock`.
+
+    Args:
+        deep_branch: Input from the previous pyramid.
+        shallow_branch: Input from the current level.
```
pmeier, line 284: The branches are modules and not images.

```suggestion
        deep_branch: Previous pyramid level.
        shallow_branch: Current pyramid level.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -224,6 +324,36 @@ def level(
...
+    A level contains either only a
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.ConvSequence` if it is the first
+    level or a :class:`~pystiche_paper.ulyanov_et_al_2016._modules.BranchBlock`
+    followed by a :class:`~pystiche_paper.ulyanov_et_al_2016._modules.ConvSequence`. The
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.BranchBlock` gets
+    ``prev_level_block`` as ``intermediate`` in an
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.HourGlassBlock` as
+    ``deep_branch`` and a
+    :class:`~pystiche_paper.ulyanov_et_al_2016._modules.ConvSequence` as
+    ``shallow_branch``.
```
pmeier, line 338: Too detailed. It should only state that

- level is a conv sequence
- if its not the first, the previous is put into hourglass block and joined with a branch block and finally followed by another conv sequence

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -224,6 +324,36 @@ def level(
...
+
+    Args:
+        prev_level_block: Optional Input from the previous level. If ``None``, only one
```
pmeier, line 341:
```suggestion
        prev_level_block: Previous pyramid level. If given, it is incorporated in the current level.
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -49,6 +56,20 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    This block comprises a downsample to half the size using
+    :class:`~torch.nn.AvgPool2d` followed by a processing with ``intermediate`` and an
+    upsample to twice the size using :class:`~torch.nn.Upsample`.
```
pmeier, line 63: Could you link to `downsample` and `upsample`?

docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst
```diff
@@ -84,7 +84,6 @@ The ``layers`` have either configuration ``1``
 | ``style_layers``          | ``1``       | ``2``               |
 +---------------------------+-------------+---------------------+
 
-
```
pmeier, line 87: Any reason to do this? If not, please revert.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -49,6 +56,22 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    This block comprises a
+    :func:`~pystiche_paper.ulyanov_et_al_2016._modules.downsample`  to half the size
+    using :class:`~torch.nn.AvgPool2d` followed by a processing with ``intermediate``
+    and an :func:`~pystiche_paper.ulyanov_et_al_2016._modules.upsample` to twice the
```
pmeier, line 64: If the function is defined within the same module you can use relative links.
```suggestion
    :func:`downsample`  to half the size
    using :class:`~torch.nn.AvgPool2d` followed by a processing with ``intermediate``
    and an :func:`upsample` to twice the
```

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -49,6 +56,22 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    This block comprises a
+    :func:`~pystiche_paper.ulyanov_et_al_2016._modules.downsample`  to half the size
+    using :class:`~torch.nn.AvgPool2d` followed by a processing with ``intermediate``
```
pmeier, line 63: How the down- and upsampling is done should be explained in `downsample` and `upsample` respectively.

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -49,6 +56,22 @@ def upsample() -> nn.Upsample:
 
 
 class HourGlassBlock(SequentialWithOutChannels):
+    r"""HourGlassBlock from :cite:`ULVL2016`.
+
+    This block comprises a
```
pmeier, line 61: I'm not happy with the text, but I can't pinpoint what it is. What do you think of "This block embeds an ``intermediate`` module between a downsample and upsample operation"?

pmeier (approved):
Almost done :wink:

PR #161, opened 2020-08-18
Add sanakoyeu et_al_2018 utils

jbueltemeier:


pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+        params: Parameter to be optimized.
+
+    """
+    if isinstance(params, nn.Module):
+        params = params.parameters()
+    return optim.Adam(params, lr=2e-4)
+
+
+def lr_scheduler(optimizer: Optimizer) -> Optional[ExponentialLR]:
```
pmeier, line 27: Why `Optional`? You are always returning an `ExponentialLR`, don't you?

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    r"""Learning rate scheduler from :cite:`SKL+2018`.
+
+    Args:
+        optimizer: Wrapped optimizer.
+
+    """
+    return DelayedExponentialLR(optimizer, gamma=0.1, delay=2)
+
+
+class ExponentialMovingAverage(FloatMeter):
```
pmeier, line 37:
```suggestion
class ExponentialMovingAverageMeter(FloatMeter):
```

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(
+        self,
+        name: str,
+        init_val: float = 0.8,
```
pmeier, line 41: Why does this start at such an odd value? `FloatMeter` starts at `0.0`.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+            val = val.item()
+        return (
+            self.last_val * (1.0 - self.smoothing_factor) + self.smoothing_factor * val
+        )
+
+    def update(self, new_val: Union[torch.Tensor, float]) -> None:  # type: ignore[override]
+        super().update(self.calculate_val(new_val))
+
+    def local_avg(self) -> float:
+        return self.last_val
```
pmeier, line 61: How is `last_val` an average of some kind?

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        val = format(self.last_val)
+        info = f"{val}"
+        return f"{self.name} {info}"
+
+
+class ContentOperatorContainer(OperatorContainer):
+    def set_target_image(self, image: torch.Tensor) -> None:
+        for op in self.children():
+            if isinstance(op, ComparisonOperator):
+                op.set_target_image(image)
```
pmeier, line 76: What is this for? If I read this right, this is now built-in since in https://github.com/pmeier/pystiche/commit/ab71e6f341cfa523a4a2ef12519577e4343be7d0

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class DelayedExponentialLR(
+    ExponentialLR
+):  # TODO: move to optim.lr_scheduler? used in two paper
```
pmeier, line 81: You can put this in `pystiche_papers.data.utils`

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
jbueltemeier: This `ExponentialMovingAverageMeter` is used to track the `discriminator_success`, which is initialized with `0.8`. Since this `ExponentialMovingAverageMeter` is not used in any other paper so far, I set the initial value to `0.8`.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
jbueltemeier: The `ExponentialMovingAverageMeter` tracks the course of the value. Therefore, the `last_val` corresponds to the current average value calculated in `calculate_val` in each `update`.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
jbueltemeier: It is for the `perceptual_loss`. I think that I introduced that because it was not like that at the time. I'm removing this one.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
pmeier: I wouldn't do that. The name does not imply that it is specifically used in this paper and thus the defaults should be general.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
pmeier: I see. But then, isn't it a `global_avg` since it includes all values ever recorded?

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
jbueltemeier: However, the Exponential Moving Average needs a starting point. Since this calculates the exponential average. So either the first value would have to be set as default or like here a user defined value. So I would suggest to leave out the default value here, so that a value has to be specified when using the `ExponentialMovingAverageMeter`.

pystiche_papers/data/utils.py
```diff
@@ -39,3 +42,32 @@ def __iter__(self) -> Iterator[Tuple[int, ...]]:
...
         return self.num_batches
+
+
+class DelayedExponentialLR(ExponentialLR):
+    r"""Decays the learning rate of each parameter group by gamma after the delay.
+
+    Args:
+        optimizer: Wrapped optimizer.
+        gamma: Multiplicative factor of learning rate decay.
+        delay: Number of epochs before the learning rate is reduced with each epoch.
```
pmeier, line 53:
```suggestion
        delay: Number of epochs before the learning rate decays.
```

pystiche_papers/data/utils.py
```diff
@@ -39,3 +42,32 @@ def __iter__(self) -> Iterator[Tuple[int, ...]]:
...
+        **kwargs: Optional parameters for the
+             :class:`~torch.optim.lr_scheduler.ExponentialLR`.
+    """
+
+    last_epoch: int
+    gamma: float
+    base_lrs: List[float]
```
pmeier, line 60: I've recently learned that this annotates class attributes, but these are instance attributes. Thus, we should move this into `__init__` with `self.last_epoch: int` etc.

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+    """
+    return DelayedExponentialLR(optimizer, gamma=0.1, delay=2)
+
+
+class ExponentialMovingAverageMeter(FloatMeter):
+    def __init__(
+        self,
+        name: str,
+        init_val: float,
```
pmeier, line 41: Is there a reason why we can't simply start with `0.0` as default?

tests/unit/data/test_utils.py
```diff
@@ -37,3 +40,26 @@ def test_InfiniteCycleBatchSampler_len():
...
     batch_sampler = utils.FiniteCycleBatchSampler(data_source, num_batches)
     assert len(batch_sampler) == num_batches
+
+
+def test_DelayedExponentialLR():
+    base_lr = 1e3
+    transformer = nn.Conv2d(3, 3, 1)
+    gamma = 0.1
+    delay = 2
+    num_steps = 5
```
pmeier, line 50: Nit

```suggestion
    num_epochs = 5
```

pystiche_papers/sanakoyeu_et_al_2018/_utils.py
```diff
@@ -0,0 +1,69 @@
...
+        if isinstance(val, torch.Tensor):
+            val = val.item()
+        return (
+            self.last_val * (1.0 - self.smoothing_factor) + self.smoothing_factor * val
+        )
+
+    def update(self, new_val: Union[torch.Tensor, float]) -> None:  # type: ignore[override]
+        super().update(self.calculate_val(new_val))
+
+    def global_avg(self) -> float:
```
pmeier, line 60:
```suggestion
    @property
    def global_avg(self) -> float:
```

tests/unit/sanakoyeu_et_al_2018/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    init_val = 0.8
+    smoothing_factor = 0.05
+    name = "test_ema"
+    ema = paper.ExponentialMovingAverageMeter(name, init_val)
+
+    with subtests.test("name"):
+        assert ema.name == name
+
+    with subtests.test("init_val"):
+        assert ema.global_avg() == init_val
```
pmeier, line 69: Please do not compare `float`s for equality.
```suggestion
        assert ema.global_avg() == pytest.approx(init_val)
```

tests/unit/sanakoyeu_et_al_2018/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_ExponentialMovingAverageMeter_str_smoke():
+    init_val = 0.8
+    meter = paper.ExponentialMovingAverageMeter(
+        "test_exponential_moving_average_meter", init_val
+    )
+    assert isinstance(str(meter), str)
+    meter.update(0.0)
+    assert isinstance(str(meter), str)
```
pmeier, line 88: Why are you testing for two steps? There is no behavior change, or is there?

tests/unit/sanakoyeu_et_al_2018/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    name = "test_ema"
+    ema = paper.ExponentialMovingAverageMeter(name, init_val)
+
+    with subtests.test("name"):
+        assert ema.name == name
+
+    with subtests.test("init_val"):
+        assert ema.global_avg() == init_val
+
+    with subtests.test("local average"):
```
pmeier, line 71:
```suggestion
    with subtests.test("global_avg"):
```

tests/unit/sanakoyeu_et_al_2018/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    vals = [0.6, 0.7, 0.8]
+    init_val = 0.8
+    smoothing_factor = 0.05
+    name = "test_ema"
+    ema = paper.ExponentialMovingAverageMeter(name, init_val)
+
+    with subtests.test("name"):
+        assert ema.name == name
+
+    with subtests.test("local average"):
```
pmeier, line 48:
```suggestion
    with subtests.test("global_avg"):
```

tests/unit/sanakoyeu_et_al_2018/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+def test_ExponentialMovingAverageMeter_update_with_tensor(subtests):
+    vals = [torch.tensor([0.6]), torch.tensor([0.7]), torch.tensor([0.8])]
+    init_val = 0.8
+    smoothing_factor = 0.05
+    name = "test_ema"
+    ema = paper.ExponentialMovingAverageMeter(name, init_val)
+
+    with subtests.test("name"):
+        assert ema.name == name
```
pmeier, line 66: You don't need to retest this, since updating with a `torch.Tensor` should have no effect on the name.
```suggestion
```

pmeier (changes requested):

pmeier (approved):

PR #162, opened 2020-08-20
Add sanakoyeu_et_al_2018 data

jbueltemeier:


tests/unit/sanakoyeu_et_al_2018/test_data.py
```diff
@@ -0,0 +1,118 @@
...
+from pystiche.data import ImageFolderDataset
+from pystiche.image import transforms
+from pystiche_papers import utils
+from pystiche_papers.data.utils import FiniteCycleBatchSampler
+from pystiche_papers.utils import make_reproducible
+
+
+def test_image_transform():
+    edge_size = 16
+    make_reproducible()
```
pmeier, line 16:
```suggestion
    utils.make_reproducible()
```

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,129 @@
...
+def batch_sampler(
+    data_source: Sized,
+    impl_params: bool = True,
+    num_batches: Optional[int] = None,
+    batch_size: Optional[int] = None,
+) -> FiniteCycleBatchSampler:
+
+    if num_batches is None:
+        # https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/main.py#L68
+        num_batches = 300_000 if impl_params else 100_000
```
pmeier, line 101: I recall something related to this. I think we split the `num_batches` in multiple epochs? Do you remember? :smile:

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,129 @@
...
+
+    transforms = (
+        OptionalRescale(),
+        ValidRandomCrop((edge_size, edge_size)),
+        OptionalGrayscaleToFakegrayscale(),
+    )
+    return ComposedTransform(*transforms)
+
+
+# TODO: Implement this
```
pmeier, line 58: Yeah, that is on my list from some time. I'll take this up for real now.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,129 @@
...
+        def __init__(
+            self, interpolation_mode: str = "bilinear",
+        ):
+            super().__init__()
+            self.interpolation_mode = interpolation_mode
+
+        def forward(
+            self, image: torch.Tensor, maximal_size: int = 1800, minimal_size: int = 800
+        ) -> torch.Tensor:
+            if max(image.shape) > maximal_size:
```
pmeier, line 27: Please correct me if I'm wrong:

- If the image is larger than `maximal_size` or smaller than `minimal_size`, it is resized to that.
- Special case: if the image is smaller than four times the `minimal_size` it is resized to the `minimal_size` on both edges

Just making sure, that I understand it correctly before reviewing.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,129 @@
...
+    class OptionalRescale(Transform):
+        # https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/prepare_dataset.py#L49-L68
+        def __init__(
+            self, interpolation_mode: str = "bilinear",
+        ):
+            super().__init__()
+            self.interpolation_mode = interpolation_mode
+
+        def forward(
+            self, image: torch.Tensor, maximal_size: int = 1800, minimal_size: int = 800
```
pmeier, line 25: Shouldn't `maximal_size` and `minimal_size` instance parameter or do they change in between calls?

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,129 @@
...
+from pystiche.image.transforms import ComposedTransform, Transform, ValidRandomCrop
+from pystiche.image.transforms import functional as F
+from pystiche.image.utils import extract_image_size
+
+from ..data.utils import FiniteCycleBatchSampler
+from ..utils import OptionalGrayscaleToFakegrayscale
+
+
+def image_transform(edge_size: int = 768,) -> ComposedTransform:
+    class OptionalRescale(Transform):
```
pmeier, line 16: I think something along the lines `BoundedRescale` would be a better fit to describe what is going on. It is not ideal, but I can't think of something better right now. `OptionalRescale` conveys that the rescaling is always optional while `BoundedRescale` conveys that the rescaling within bounds always happens. Ideally, we need something that combines these two.

tests/unit/sanakoyeu_et_al_2018/test_data.py
```diff
@@ -0,0 +1,118 @@
...
+import torch
+from torch.utils.data import DataLoader
+
+import pystiche.image.transforms.functional as F
+import pystiche_papers.sanakoyeu_et_al_2018 as paper
+from pystiche.data import ImageFolderDataset
+from pystiche.image import transforms
+from pystiche_papers import utils
+from pystiche_papers.data.utils import FiniteCycleBatchSampler
+from pystiche_papers.utils import make_reproducible
```
pmeier, line 11:
```suggestion
```

pystiche_papers/sanakoyeu_et_al_2018/_data.py
pmeier, line 16: Since this has multiple branches, I suggest we move it outside of `image_transform` to make the testing easier.

tests/unit/sanakoyeu_et_al_2018/test_data.py
```diff
@@ -0,0 +1,118 @@
...
+    transform = transforms.ValidRandomCrop(edge_size)
+    utils.make_reproducible()
+    desired = F.grayscale_to_fakegrayscale(transform(image))
+
+    ptu.assert_allclose(actual, desired)
+
+
+def test_image_transform_optional_rescale(subtests):
+
+    with subtests.test("too large image"):
```
pmeier, line 32: Since no data is shared between three subtests, there is no need for subtests here. Please split this into three separate tests.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def style_dataset(
+    root: str,
+    style: str,
+    transform: Optional[nn.Module] = None,
+    download: bool = False,
+) -> WikiArt:
+    if transform is None:
+        transform = image_transform()
```
pmeier, line 137: @jbueltemeier Is the image transform used for content and style images alike?

pmeier (commented):
I'll add the content dataset as well as tests later.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
jbueltemeier: Yeah, that's right.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
jbueltemeier: The transformation of the images as it is now is used for the content and style. However, I just saw that the images from the places dataset are enlarged to twice the size before the previous transformation is applied. I'll integrate that.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
jbueltemeier: That's right. The paper describes that the `lr` is reduced by a factor of 10 after 200k steps.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
pmeier: In that case, why are you using `Rescale` at all? I think `Resize` would be the better fit here. As a rule of thumb: If you the size you want to resize to, use `Resize`. If the size you want to resize to is defined relative to the original size, i.e. double or half, use `Rescale`.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
pmeier: According to this without `impl_params` we only have 100k steps. Thus, both can't be true. Which one is it?

pystiche_papers/sanakoyeu_et_al_2018/_data.py
pmeier: What about `ClampSize`?

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,224 @@
...
+        # The num_iterations are split up into multiple epochs with corresponding
+        # num_batches:
+        # The number of epochs is defined in _nst.training .
+        # 300_000 = 1 * 300_000
+        # https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/main.py#L68
+        # 300_000 = 3 * 100_000
+        num_batches = 300_000 if impl_params else 100_000
+
+    if batch_size is None:
+        batch_size = 1
```
pmeier, line 199: Is this an artifact or is this WIP?

pmeier (changes requested):

jbueltemeier:
There is still an image augmentor missing which modifies the randomly selected image from the dataset by some transformations like scaling or rotation. See [here](https://github.com/CompVis/adaptive-style-transfer/blob/master/img_augm.py). I don't think that this is a part of the `image_transform`, but one that belongs to the `batch_sampler`. Or what do you think?

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -0,0 +1,218 @@
...
+        dct = super()._properties()
+        dct["min_edge_size"] = self.min_edge_size
+        dct["max_edge_size"] = self.max_edge_size
+        if self.interpolation_mode != "bilinear":
+            dct["interpolation_mode"] = self.interpolation_mode
+        return dct
+
+
+def image_transform(edge_size: int = 768,) -> transforms.ComposedTransform:
+    # TODO: Add image augmentator with random affine transform hsv transfomration and
```
pmeier, line 87: Could you list all needed transformations with their respective parameters? I'll handle it from there.

pmeier (approved):

pmeier:
@jbueltemeier During rebase, I accidentally removed the commits you made to this branch today. Could you re-add them?

pmeier:
@jbueltemeier The augmentation proved to be a lot of more work than I expected. I'll merge this now and add them in a follow-up PR.

pmeier:
Superseded by #177.

PR #166, opened 2020-08-25
Sanakoyeu 2018 basic modules

jbueltemeier:


pmeier:
Its seems you used a outdated base for your branch. Could you rebase to the current HEAD of `sanakoyeu-2018`? Otherwise coverage reports will probably not work.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+    return nn.Conv2d(
+        in_channels, out_channels, kernel_size, stride=stride, padding=padding
+    )
+
+
+class ConvBlock(nn.Sequential):
+    r"""ConvBlock from :cite:`SKL+2018`.
+
+    This block comprises a :class:`~torch.nn.Conv2d` followed by normalization with
+    :class:`~torch.nn.InstanceNorm2d` and if ``act is not None`` an activation.
```
pmeier, line 55: Too detailed.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+
+def get_padding(
+    padding: str, kernel_size: Union[Tuple[int, int], int]
+) -> Union[Tuple[int, int], int]:
+    if padding == "same":
+        return cast(Tuple[int, int], same_size_padding(kernel_size))
+    elif padding == "valid":
+        return 0
+    else:
+        raise ValueError
```
pmeier, line 26: Please add a descriptive error message. You could use `pystiche.misc.verify_str_arg`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def activation(act: str = "relu", inplace: bool = True) -> nn.Module:
+    if act == "relu":
+        return nn.ReLU(inplace=inplace)
+    elif act == "lrelu":
+        return nn.LeakyReLU(negative_slope=0.2, inplace=inplace)
+    else:
+        raise ValueError
```
pmeier, line 35: Same as above.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+    "get_padding",
+    "activation",
+    "conv",
+    "ConvBlock",
+    "ConvTransponseBlock",
+    "residual_block",
+]
+
+
+def get_padding(
```
pmeier, line 18: Why is this called `get_padding` but the function below `activation` (without the `get`)? We should keep this consistent.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+            modules.append(activation(act=act, inplace=inplace))
+
+        super().__init__(*modules)
+
+
+class ConvTransponseBlock(nn.Module):
+    r"""ConvTransponse from :cite:`SKL+2018`.
+
+    This block comprises a :func:`~torch.nn.functional.interpolate` followed by a
+    :class:`~pystiche_paper.sanakoyeu_et_al_2018._modules.ConvBlock`.
```
pmeier, line 106: Too detailed. You do not need to explain what the internals are, but rather what they do.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+    :class:`~torch.nn.InstanceNorm2d` and if ``act is not None`` an activation.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels: Number of channels produced by the convolution.
+        kernel_size: Size of the convolving kernel.
+        stride: Stride of the convolution. Defaults to ``1``.
+        padding: Padding of the input. It can be either ``"valid"`` for no padding or
+            ``"same"`` for padding to preserve the size. Defaults to ``"valid"``.
+        act: The act is either ``"relu"`` for a :class:`~torch.nn.ReLU`, ``"lrelu"`` for
```
pmeier, line 64:
```suggestion
        act: The activation is either ``"relu"`` for a :class:`~torch.nn.ReLU`, ``"lrelu"`` for
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+    :class:`~pystiche_paper.sanakoyeu_et_al_2018._modules.ConvBlock`.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels: Number of channels produced by the convolution.
+        kernel_size: Size of the convolving kernel.
+        stride: Stride of the interpolation. Defaults to ``2``.
+        padding: Padding of the input. It can be either ``"valid"`` for no padding or
+            ``"same"`` for padding to preserve the size. Defaults to ``"same"``.
+        act: The act is either ``"relu"`` for a :class:`~torch.nn.ReLU`, ``"lrelu"`` for
```
pmeier, line 115:
```suggestion
        act: The activation is either ``"relu"`` for a :class:`~torch.nn.ReLU`, ``"lrelu"`` for
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+            self.conv(
+                nn.functional.interpolate(
+                    input, scale_factor=self.stride, mode="nearest"
+                )
+            ),
+        )
+
+
+def residual_block(channels: int) -> ResidualBlock:
+    r"""Residualblock from :cite:`SKL+2018`.
```
pmeier, line 161:
```suggestion
    r"""Residual block from :cite:`SKL+2018`.
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,182 @@
...
+
```
pmeier, line 162: Description is missing.

pmeier (approved):

pmeier (changes requested):
Approval was a miss click.

tests/unit/sanakoyeu_et_al_2018/test_modules.py
```diff
@@ -0,0 +1,127 @@
...
+    ):
+        with subtests.test(str_padding):
+            actual = paper.get_padding(str_padding, kernel_size)
+            assert actual == desired
+
+
+def test_get_padding_wrong_string():
+    kernel_size = 3
+    with pytest.raises(ValueError):
+        paper.get_padding("", kernel_size)
```
pmeier, line 25: You don't need to test this, since the behavior is handled in `pystiche.misc.verify_str_arg`. As a rule of thumb you only need to test `Exception`s that are raised by our code.

tests/unit/sanakoyeu_et_al_2018/test_modules.py
```diff
@@ -0,0 +1,127 @@
...
+                assert actual.inplace
+
+            if isinstance(actual, nn.LeakyReLU):
+                with subtests.test("slope"):
+                    assert actual.negative_slope == pytest.approx(0.2)
+
+
+def test_get_activation_wrong_string():
+    with pytest.raises(ValueError):
+        paper.get_activation("")
```
pmeier, line 45: Same as above.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,185 @@
...
+        modules.append(nn.InstanceNorm2d(out_channels))
+
+        if act is not None:
+            modules.append(get_activation(act=act, inplace=inplace))
+
+        super().__init__(*modules)
+
+
+class ConvTransponseBlock(nn.Module):
```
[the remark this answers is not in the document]
jbueltemeier: You got that right. The authors also use the term `deconv2d` for interpolation followed by a convolution so I would change the name to `DeConvBlock`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,186 @@
...
+
+        modules.append(nn.InstanceNorm2d(out_channels))
+
+        if act is not None:
+            modules.append(get_activation(act=act, inplace=inplace))
+
+        super().__init__(*modules)
+
+
+class DeConvBlock(nn.Module):
```
pmeier, line 102: In the paper they call it "upsampling block". Do they use `DeConv` in the implementation? If not, I suggest we use `UpsampleConvBlock`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,186 @@
...
+    r"""DeConvBlock from :cite:`SKL+2018`.
+
+    This block upsamples the input to twice the size followed by a :class:`ConvBlock`.
```
pmeier, line 105: Twice is not true here, is it? The upsampling depends on stride.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: They call it [deconv2d](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/ops.py#L58) in the implementation.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: Yes it depends on `stride`, but this block is only used with a `stride=2`. This should also be renamed, since the `stride` here is only used as `scaling_factor`.

pmeier: In that case, go ahead with the rename.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: The problem I see is that de-convolution is an [existing term](https://en.wikipedia.org/wiki/Deconvolution) that basically describes the inverse operation of a given convolution. Although the authors used that in their reference implementation, I think we should not use this to avoid confusion.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,186 @@
...
+    * a tuple of two :class:`int` s – in which case, the first int is used for the
+      vertical dimension, and the second int for the horizontal dimension
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: Union[Tuple[int, int], int],
+        scale_factor: Union[Tuple[int, int], int] = 2,
```
pmeier, line 131:
```suggestion
        scale_factor: Union[Tuple[float, float], float] = 2.0,
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -0,0 +1,186 @@
...
+class UpsampleConvBlock(nn.Module):
+    r"""UpsampleConvBlock from :cite:`SKL+2018`.
+
+    This block upsamples the input followed by a :class:`ConvBlock`.
+
+    Args:
+        in_channels: Number of channels in the input.
+        out_channels: Number of channels produced by the convolution.
+        kernel_size: Size of the convolving kernel.
+        scale_factor: ``scale_factor`` of the interpolation. Defaults to ``2``.
```
pmeier, line 111:
```suggestion
        scale_factor: ``scale_factor`` of the interpolation. Defaults to ``2.0``.
```

pmeier (approved):

PR #171, opened 2020-08-26
Sanakoyeu 2018 transformer modules

jbueltemeier:


pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
     )
 
     return ResidualBlock(residual)
+
+
+def transformer_encoder(in_channels: int = 3,) -> SequentialEncoder:
```
pmeier, line 196: Can't we just call this `encoder`?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+        ConvBlock(in_channels=in_channels, out_channels=32, kernel_size=3, stride=1),
+        ConvBlock(in_channels=32, out_channels=32, kernel_size=3, stride=2),
+        ConvBlock(in_channels=32, out_channels=64, kernel_size=3, stride=2),
+        ConvBlock(in_channels=64, out_channels=128, kernel_size=3, stride=2),
+        ConvBlock(in_channels=128, out_channels=256, kernel_size=3, stride=2),
+    )
+    return SequentialEncoder(modules)
+
+
+def transformer_decoder(
```
pmeier, line 214: Same here. IMO simply `decoder` should be enough.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    out_channels: int = 3, num_res_block: int = 9,
```
pmeier, line 215:
```suggestion
    out_channels: int = 3, num_residual_blocks: int = 9,
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    )
+    return pystiche.SequentialModule(*modules)
+
+
+class DecoderSigmoidOutput(nn.Module):
+    def __init__(self) -> None:
+        super().__init__()
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        return torch.sigmoid(input) * 2 - 1
```
pmeier, line 244: So simply `torch.tanh(input / 2)`?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+class Decoder(nn.Module):
+    r"""Decoder part of the :class:`Transformer` from :cite:`SKL+2018`."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.decoder = transformer_decoder()
+        self.output_module = DecoderSigmoidOutput()
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        return cast(torch.Tensor, self.output_module(self.decoder(input)))
```
pmeier, line 256: Are you sure about this? `DecoderSigmoidOutput` maps the values to the interval `(-1.0, 1.0)`. Is there some other kind of post-processing?

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: Yes, there is still a pre- and postprocessing of the images  to the interval `(-1.0, 1.0)`.

pmeier: In that case, please implement this in `_utils.py`

jbueltemeier: I will do this in a following PR.

pmeier: Please do that here, since it is directly related.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+
+def encoder(in_channels: int = 3,) -> SequentialEncoder:
+    r"""Encoder part of the :class:`Transformer` from :cite:`SKL+2018`.
+
+    Args:
+        in_channels: Number of channels in the input. Defaults to ``3``.
+
+    """
+    modules = (
```
pmeier, line 203: This seems like my comment from https://github.com/pmeier/pystiche_papers/pull/172#discussion_r477273839 applies here as well. To make this a little simpler, we can use `more_itertools.pairwise` instead of writing our own function. To add `more_itertools` as a requirement, add it to 

https://github.com/pmeier/pystiche_papers/blob/0b8dfbd0eb67c0f40059c28a0edd82a34c0e0581/setup.cfg#L27-L30

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+def decoder(
+    out_channels: int = 3, num_residual_blocks: int = 9,
+) -> pystiche.SequentialModule:
+
+    modules: List[nn.Module] = []
+    for i in range(num_residual_blocks):
+        modules.append(residual_block(256))
+
+    modules.append(UpsampleConvBlock(in_channels=256, out_channels=256, kernel_size=3))
```
pmeier, line 222: Same here.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,78 @@ def residual_block(channels: int) -> ResidualBlock:
...
+            out_channels=out_channels,
+            kernel_size=7,
+            stride=1,
+            padding="valid",
+        )
+    )
+    return pystiche.SequentialModule(*modules)
+
+
+class DecoderTanhOutput(nn.Module):
```
pmeier, line 239: In `johnson_alahi_li_2016` we called this 

https://github.com/pmeier/pystiche_papers/blob/0b8dfbd0eb67c0f40059c28a0edd82a34c0e0581/pystiche_papers/johnson_alahi_li_2016/_modules.py#L248-L250

While I'm not super happy with the name I think its better than `DecoderTanhOutput`. The former describes the intent while the latter the content.

pmeier, line 239: I think its OK to do this as a local class in `Decoder` since it is useless otherwise.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,91 @@ def residual_block(channels: int) -> ResidualBlock:
...
+        in_channels: Number of channels in the input. Defaults to ``3``.
+
+    """
+    modules = [
+        nn.ReflectionPad2d(15),
+        ConvBlock(in_channels=in_channels, out_channels=32, kernel_size=3, stride=1),
+    ]
+    channels = (32, 32, 64, 128, 256)
+
+    modules.extend(
```
pmeier, line 210: I will be even cleaner if you use `conv_block = functools.partial`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,91 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.decoder = decoder()
+        self.output_module = self.ValueRangeDelimiter()
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        return cast(torch.Tensor, self.output_module(self.decoder(input)))
+
+    class ValueRangeDelimiter(nn.Module):
```
pmeier, line 266: To take this even further: You can define this inside `__init__`. If you do, its no accessible afterwards, which is preferable here.

setup.cfg
```diff
@@ -28,6 +28,7 @@ install_requires =
     # TODO: add a released version of pystiche
     torch >= 1.6
     torchvision >= 0.7
+    more_itertools >= 8.4
```
pmeier, line 31: Why the constraint to `>=8.4`. Is this the first release where `pairwise` is included?

pmeier (approved):
Only minor changes left.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +193,87 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+class Decoder(nn.Module):
+    r"""Decoder part of the :class:`Transformer` from :cite:`SKL+2018`."""
+
+    def __init__(self) -> None:
+        class ValueRangeDelimiter(nn.Module):
+            r"""Maps the values to the interval (-1.0, 1.0)."""
+
+            def forward(self, input: torch.Tensor) -> torch.Tensor:
+                return torch.tanh(input / 2)
```
pmeier, line 260: This makes the representation of `Decoder` more clear.

```suggestion
                return torch.tanh(input / 2)
                
            def _properties() -> Dict[str, Any]:
                dct = super()._properties()
                dct["output_range"] = "(-1.0, 1.0)"
                return dct
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +193,87 @@ def residual_block(channels: int) -> ResidualBlock:
...
+            out_channels=out_channels,
+            kernel_size=7,
+            stride=1,
+            padding="valid",
+        )
+    )
+    return pystiche.SequentialModule(*modules)
+
+
+class Decoder(nn.Module):
```
pmeier, line 252: This is needed for the properties handling.

```suggestion
class Decoder(pystiche.Module):
```

pmeier, line 252: Better yet, can't we move this into `decoder`? I don't think we need a separate module for the `tanh`.

pmeier (changes requested):

pmeier:
@jbueltemeier I've rebased, refactored and fixed a little here and there :smirk:

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +190,75 @@ def residual_block(channels: int) -> ResidualBlock:
...
+            out_channels=out_channels,
+            kernel_size=7,
+            stride=1,
+            padding="valid",
+        ),
+        ValueRangeDelimiter(),
+    )
+
+
+class Transformer(nn.Module):
```
pmeier, line 255: Could you add a function `transformer` that returns an instance of this in compliance with the others parts?

pystiche_papers/utils/misc.py
```diff
@@ -253,3 +255,14 @@ def load_state_dict_from_url(
         return cast(
             Dict[str, torch.Tensor], torch.load(cached_file, map_location=map_location)
         )
+
+
+T = TypeVar("T")
+
+
+def channel_progression(
```
pmeier, line 263: Could you add tests for this? If this is tested, you may also use this from other tests to make them more readable.

pmeier (changes requested):

pmeier (approved):

PR #172, opened 2020-08-26
Add Sanakoyeu 2018 Discriminator

jbueltemeier:


pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+def get_transformation_block(
+    in_channels: int,
+    kernel_size: Union[Tuple[int, int], int],
+    stride: Union[Tuple[int, int], int],
+    padding: Union[Tuple[int, int], int],
+    impl_params: bool = True,
+) -> nn.Module:
+    if impl_params:
+        return nn.AvgPool2d(kernel_size=kernel_size, stride=stride, padding=padding)
+    return nn.Conv2d(in_channels, 3, kernel_size, stride=stride, padding=padding)
```
pmeier, line 206: Are you sure about this? `Conv2d` and `AvgPool2d` are quite different blocks. If this is the case please add the link to the reference implementation.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+        self.forwardBlock = get_transformation_block(
+            in_channels, kernel_size, stride, padding, impl_params=impl_params,
+        )
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if self.impl_params:
+            return cast(torch.Tensor, self.forwardBlock(input))
+        else:
+            return cast(torch.Tensor, nn.utils.weight_norm(self.forwardBlock(input)))
```
pmeier, line 249: Never worked with `weight_norm` before, but from the [documentation](https://pytorch.org/docs/stable/generated/torch.nn.utils.weight_norm.html#torch.nn.utils.weight_norm) I think this has to be applied only once. Could you check?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+
+def discriminator_encoder_modules(
+    in_channels: int = 3, inplace: bool = True,
+) -> Dict[str, nn.Sequential]:
+    # FIXME:  if/when the Python interpreter will learn to accept the correct signature
+    # https://stackoverflow.com/questions/41207128/how-do-i-specify-ordereddict-k-v-types-for-mypy-type-annotation
```
pmeier, line 256: Yeah, this is only possible for Python `>=3.7`. Lets ignore all errors for now and simply remove the ignores as soon as we bump the minimal required python version.

```suggestion
    # FIXME:  if/when the Python interpreter will learn to accept the correct signature
    # https://stackoverflow.com/questions/41207128/how-do-i-specify-ordereddict-k-v-types-for-mypy-type-annotation
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+    modules = OrderedDict(
+        {
```
pmeier, line 259: Never initialize an `OrderedDict` with a `dict` if you care about the order. Since `dict` does not keep the items in order they might be already unordered before they are passed to `OrderedDict`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+                1024,
+                kernel_size=5,
+                stride=2,
+                padding="same",
+                act="lrelu",
+                inplace=inplace,
+            ),
+        }
+    )
+    return cast(Dict[str, nn.Sequential], modules)
```
pmeier, line 325: This can be made a lot more legible with a little syntactic sugar

```suggestion
from functools import partial
import itertools
from typing import Iterable


def pairwise(iterable: Iterable):
    a, b = itertools.tee(iterable)
    next(b, None)
    return zip(a, b)


def discriminator_encoder_modules(in_channels: int = 3, inplace: bool = True):
    conv_block = partial(
        ConvBlock, kernel_size=5, stride=2, padding="same", act="lrelu", inplace=inplace
    )

    channels = (in_channels, 128, 128, 256, 512, 512, 1024)
    return OrderedDict(
        [
            (f"scale_{idx}", conv_block(in_channels, out_channels))
            for idx, (in_channels, out_channels) in pairwise(channels)
        ]
    )
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    """
+
+    def __init__(self, in_channels: int = 3) -> None:
+        super().__init__(
+            self._collect_modules(
+                discriminator_encoder_modules(in_channels=in_channels)
+            )
+        )
+
+    def _collect_modules(
```
pmeier, line 342: I'm confused: why are you first naming the blocks `scale_x` just to rename them here?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+                padding="same",
+                act="lrelu",
+                inplace=inplace,
+            ),
+        }
+    )
+    return cast(Dict[str, nn.Sequential], modules)
+
+
+class DiscriminatorEncoder(MultiLayerEncoder):
```
pmeier, line 328: If there is an "encoder part" shouldn't there also be a `Decoder`?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +192,202 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    return conv(
+        in_channels=in_channels,
+        out_channels=1,
+        kernel_size=kernel_size,
+        stride=1,
+        padding=padding,
+    )
+
+
+def get_prediction_modules() -> Dict[str, nn.Module]:
```
pmeier, line 386: What would you use this for?

pmeier (changes requested):
I'm having troubles understanding the purpose of the individual parts. I'll re-read the paper, but you can answer some questions in the comments I made.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: No a `decoder` is not needed here. This `encoder `is a `multi_layer_encoder `for a modified `MultiLayerEncodingOperator`, since the `discriminator` structure has a forward part, from which the encodings are evaluated on predefined layers using the additional `prediction_modules`. I should probably rename this to a `discriminator_multi_layer_encoder` though.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: I use this later to set the default `prediction_modules` from the paper in the `discriminator_operator`. I think this will be clearer if the file _loss.py is integrated.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: The authors also describe this (See [here](https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/module.py#L238)).

> This is a simplified version of transformer block described in our paper

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: As I understood the `MultiLayerEncoder`, it does not work with a `nn.Sequential`. But since the `ConvBlock`s are a `nn.Sequence` of convolution, normalization and activation, I first named the `ConvBlock`s as in the paper with `scale_x` and then I brought it into `DiscriminatorEncoder` into the structure of the `MultiLayerEncoder`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: You are right. [source](https://discuss.pytorch.org/t/nn-utils-weight-norm-not-working-for-me/7708/3)

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: So instead of the conv block they simply use an avg pool? :smile:

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: I've never tried this, but in theory there is nothing that prevents us from using sequentials within an `MultiLayerEncoder`. If we do that each sequential is treated as a single layer, i.e. we can't access the inner encodings. But AFAIK this is not needed here, is it?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: Maybe we should split this into two entitites: `Discriminator` and `DiscriminatorMultiLayerEncoder`. Although we won't use the `Discriminator` for our implementation (other than as base for the `DiscriminatorMultiLayerEncoder`) I think it makes the distinction more clear.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: In that case, please remove this for now and only implement this when its needed.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: Only the encodings after activation are needed here. So you are right in your assumption that the inner encodings are not needed.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: Right. I suppose that the convolution has led to problems, because the parameters can be learned additionally during the training.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +194,154 @@ def residual_block(channels: int) -> ResidualBlock:
...
+        return nn.AvgPool2d(kernel_size=kernel_size, stride=stride, padding=padding)
+    return cast(
+        nn.Module,
+        nn.utils.weight_norm(
+            nn.Conv2d(in_channels, 3, kernel_size, stride=stride, padding=padding)
+        ),
+    )
+
+
+class TransformerBlock(nn.Module):
```
pmeier, line 217: Just realized this: the transformer block is also only needed for the loss calculation, correct? If so, please remove this for now to keep the scope of PR tight.

jbueltemeier: In this case, the `DiscriminatorMultiLayerEncoder` is also only required for the calculation of the loss. Therefore I would suggest at this point that I extend this PR by the necessary classes and function from loss?

pmeier: This raises the mental burden to review this, which I like to avoid. Please remove the out-of-scope parts for now and re-add them in a follow-up PR.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: The complete `Discriminator `is defined as a `MultiLayerEncodingOperator `in loss, if that's what you mean.

pmeier: > The complete `Discriminator` is defined as a `MultiLayerEncodingOperator` in loss, if that's what you mean.

That doesn't make sense. Do you mean: "The `Discriminator` servers as `MultiLayerEncoder` within a `MultiLayerEncodingOperator`"?

---

What I mean is that we should have a simple sequential `Discriminator` as the authors describe in their paper. For our implementation (to match the reference implementation) we also need an `MultiLayerDiscriminator`. I think we should keep both separate to avoid confusion.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: In that case, just use the sequentials as layers.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: Yes, i mean the `Discriminator` serves as `MultiLayerEncoder` within a `MultiLayerEncodingOperator`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +190,58 @@ def residual_block(channels: int) -> ResidualBlock:
     )
 
     return ResidualBlock(residual)
+
+
+def pairwise(iterable: Iterable) -> Iterator[Tuple[Any, Any]]:
```
pmeier, line 195: Lets wait for #171 to land and use the same methodology here.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +190,58 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    a, b = itertools.tee(iterable)
+    next(b, None)
+    return zip(a, b)
+
+
+def discriminator_encoder_modules(
```
pmeier, line 201: I think we are still talking past each other. I want a `class Discriminator(pystiche.Module)` and a `class DiscriminatorMultiLayerEncoder(enc.MultiLayerEncoder)` that is based on the former.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +190,58 @@ def residual_block(channels: int) -> ResidualBlock:
...
+    in_channels: int = 3, inplace: bool = True
+) -> List[Tuple[str, nn.Sequential]]:
+    conv_block = partial(
+        ConvBlock, kernel_size=5, stride=2, padding="same", act="lrelu", inplace=inplace
+    )
+
+    channels = (in_channels, 128, 128, 256, 512, 512, 1024, 1024)
+    return [
+        (f"scale_{idx}", conv_block(in_channels, out_channels))
```
pmeier, line 210: Does the prefix `scale_` serve any purpose? If not, wouldn't be a simple `pystiche.SequentialModule` the better fit here?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +190,58 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+    Args:
+        in_channels: Number of channels in the input. Defaults to ``3``.
+    """
+
+    def __init__(self, in_channels: int = 3) -> None:
+        super().__init__(discriminator_encoder_modules(in_channels=in_channels))
+
+
+def prediction_module(
```
pmeier, line 226: IIRC this is part of the `style_loss`, correct? If that is the case, please revert his for now and include this in the PR for the `style_loss`.

pmeier (changes requested):
Linting error should be fixed after a rebase.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: The prefix `scale_i` is used to select the scales `i` on which the `discriminator_loss` is calculated.

pmeier: Yeah, but what advantage does `f"scale_{idx}"` offer over `str(idx)` whih is the default of a sequential?

jbueltemeier: It is the same name that was chosen in the the [reference_implementation](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L134).

pmeier: If it does not serve any purpose, we should go with the simpler variant, since we aim to replicate the functionality and not each individual part.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +191,46 @@ def residual_block(channels: int) -> ResidualBlock:
...
     return ResidualBlock(residual)
+
+
+def pairwise(iterable: Iterable) -> Iterator[Tuple[Any, Any]]:
+    a, b = itertools.tee(iterable)
+    next(b, None)
+    return zip(a, b)
+
+
+def discriminator_encoder_modules(
```
pmeier, line 202: Does this function serve any purpose, i.e. is there any use case where I would want to call that outside of the `Discriminator`? If yes, could you give me an example? If not, please move this into the constructor to `Discriminator`.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -273,3 +276,45 @@ def transformer(style: Optional[str] = None,) -> Transformer:
...
+
+class Discriminator(pystiche.Module):
+    r"""Discriminator from :cite:`SKL+2018`.
+
+    Args:
+        in_channels: Number of channels in the input. Defaults to ``3``.
+    """
+
+    def __init__(self, in_channels: int = 3) -> None:
+        def discriminator_encoder_modules(
```
pmeier, line 289: What is the point of this function? it is only used once and thus the functionality could be simply inlined.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -273,3 +276,45 @@ def transformer(style: Optional[str] = None,) -> Transformer:
...
+                ConvBlock,
+                kernel_size=5,
+                stride=2,
+                padding="same",
+                act="lrelu",
+                inplace=inplace,
+            )
+
+            return channel_progression(
+                lambda in_channels, out_channels: conv_block(in_channels, out_channels),
```
pmeier, line 302: 1. `channel_progression` expects a callable with the arguments `in_channels` and `out_channels`. Thus, simply passing `conv_block` would be sufficient.
2. Since we have `channel_progression` now, `partial` is no longer needed. Simply pass a `lambda` callable of `ConvBlock`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -273,3 +276,45 @@ def transformer(style: Optional[str] = None,) -> Transformer:
...
+class DiscriminatorMultiLayerEncoder(enc.MultiLayerEncoder):
+    r"""Discriminator from :cite:`SKL+2018` as :class:`pystiche.enc.MultiLayerEncoder`.
+
+    Args:
+        in_channels: Number of channels in the input. Defaults to ``3``.
+    """
+
+    def __init__(self, in_channels: int = 3) -> None:
+        discriminator = Discriminator(in_channels=in_channels)
+        super().__init__(tuple(discriminator.named_children()))
```
pmeier, line 320:
```suggestion
        super().__init__(tuple(Discriminator(in_channels=in_channels).named_children()))
```

pmeier (changes requested):

pmeier (approved):

PR #174, opened 2020-08-28
Add sanakoyeu_et_al_2018 Transformerblock 

jbueltemeier:


pystiche_papers/sanakoyeu_et_al_2018/__init__.py
```diff
@@ -1,2 +1,3 @@
+from ._loss import transformed_image_loss
```
pmeier, line 1:
```suggestion
from ._loss import *
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -0,0 +1,34 @@
+from typing import Optional
+
+import pystiche_papers.sanakoyeu_et_al_2018 as paper
+from pystiche.enc.encoder import Encoder
+from pystiche.ops.comparison import FeatureReconstructionOperator
+
+
```
pmeier, line 7:
```suggestion
__all__ = ["transformed_image_loss"]

```

pmeier (changes requested):
The combination of `get_transformation_block` and `TransformerBlock` seems rather weird. Correct me if I'm wrong, but I think the `Transformer` has only two requirements:

1. It has to an instance of `pystiche.enc.Encoder`
2. It should apply `nn.AvgPool2d` or `nn.Conv2d` with respect to `impl_params`

If that is the case, I think we can do this much clearer:

```python
from pystiche import enc


class TransformerBlock(enc.SequentialEncoder):
    r"""TransformerBlock from :cite:`SKL+2018`.

    This block takes an image as input and produce a transformed image of the same size.

    Args:
        in_channels: Number of channels in the input. Defaults to ``3``.
        kernel_size: Size of the convolving kernel. Defaults to ``10``.
        stride: Stride of the convolution. Defaults to ``1``.
        padding: Padding of the input. It can be either ``"valid"`` for no padding or
            ``"same"`` for padding to preserve the size. Defaults to ``"same"``.
        impl_params: If ``True``, use the parameters used in the reference
            implementation of the original authors rather than what is described in
            the paper.

    If ``impl_params is True``, an :class:`~torch.nn.AvgPool2d` is used instead of a
    :class:`~torch.nn.Conv2d` with :func:`~torch.nn.utils.weight_norm`.
    """

    def __init__(
        self,
        in_channels: int = 3,
        kernel_size: Union[Tuple[int, int], int] = 10,
        stride: Union[Tuple[int, int], int] = 1,
        padding: str = "same",
        impl_params: bool = True,
    ):
        kwargs = {
            "kernel_size": kernel_size,
            "stride": stride,
            "padding": get_padding(padding, kernel_size),
        }
        module = (
            nn.AvgPool2d(**kwargs)
            if impl_params
            else nn.utils.weight_norm(nn.Conv2d(in_channels, in_channels, **kwargs))
        )
        super().__init__((module,))
```

Thoughts?

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -0,0 +1,36 @@
...
+            the paper.
+        score_weight: Score weight of the operator. If omitted, the score_weight is
+            determined with respect to ``impl_params``. Defaults to ``1e2`` if
+            ``impl_params is True`` otherwise ``1e0``.
+    """
+    if score_weight is None:
+        if impl_params:
+            score_weight = 1e2
+        else:
+            score_weight = 1e0
```
pmeier, line 31:
```suggestion
        score_weight = 1e2 if impl_params else 1e0
```

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +186,50 @@ def residual_block(channels: int) -> ResidualBlock:
...
+        padding: str = "same",
+        impl_params: bool = True,
+    ):
+        kwargs = {
+            "kernel_size": kernel_size,
+            "stride": stride,
+            "padding": get_padding(padding, kernel_size),
+        }
+        module = (
+            nn.AvgPool2d(**kwargs)  # type: ignore[arg-type]
```
pmeier, line 224: Why do you need an `ignore[arg-type]` here? What is the error message? Is it maybe enough to declare `module: nn.Module = ...`?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +186,50 @@ def residual_block(channels: int) -> ResidualBlock:
...
+            if impl_params
+            else nn.utils.weight_norm(nn.Conv2d(in_channels, in_channels, **kwargs))  # type: ignore[arg-type]
+        )
+        super().__init__((module,))
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        input = nn.functional.pad(input, [0, 1, 0, 1])
```
pmeier, line 231: Why do we need padding here? Isn't that handled by the module itself?

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -184,3 +186,50 @@ def residual_block(channels: int) -> ResidualBlock:
...
+
+        for module in self.children():
+            input = module(input)
+        return input
```
pmeier, line 235: If we need that extra padding for some reason that could be made more robust by 
```suggestion
        return super().forward(input)
```

pmeier (changes requested):
Test failures are related. Please address them in addition to my comments.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: The size of the image will be changed otherwise, but I think this should be integrated in `same_size_padding` and not here. It is probably due to the `kernel_size=10`.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
jbueltemeier: > pystiche_papers\sanakoyeu_et_al_2018\_modules.py: note: In member "__init__" of class "TransformerBlock":
pystiche_papers\sanakoyeu_et_al_2018\_modules.py:224: error: Argument 1 to "AvgPool2d" has incompatible type
"**Dict[str, Union[Tuple[int, int], int]]"; expected "bool"  [arg-type]
                nn.AvgPool2d(**kwargs)

At nn. Conv2d are similar error messages.

pmeier: Hm, OK I didn't think of that. In that case, I think it is OK as is.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
pmeier: Yeah, that means `same_size_padding` does not work as expected. Could you fix this in another PR and rebase this after the other is merged?

pmeier: Fixed in #175

jbueltemeier:
We need the `SameSizeConv2D` fixed in [#175](https://github.com/pmeier/pystiche_papers/pull/175) also for an `nn.AvgPool2D`.

pmeier:
> We need the `SameSizeConv2D` fixed in [#175](https://github.com/pmeier/pystiche_papers/pull/175) also for an `nn.AvgPool2D`.

I'm on it.

pmeier:
#179 adds the functionality that you need.

pmeier (approved):
@jbueltemeier This is good to merge after the conflicts are resolved.

PR #175, opened 2020-09-02
Fix same size padding

jbueltemeier:
As mentioned in [#174](https://github.com/pmeier/pystiche_papers/pull/174#discussion_r481221830).

If the `kernel_size` is even, the `same_size_padding` requires an additional right column at the height and/or an additional bottom column so that the `input_size `is not changed.

### DIFF
--- a/pystiche_papers/johnson_alahi_li_2016/_modules.py
+++ b/pystiche_papers/johnson_alahi_li_2016/_modules.py
@@ -1,7 +1,7 @@
 import csv
 from math import sqrt
 from os import path
-from typing import Any, Dict, List, Optional, Tuple, Union, cast
+from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
 
 import torch
 from torch import nn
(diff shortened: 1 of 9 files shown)


pmeier (commented):
The problem I see with you approach is that it is impossible to use `nn.Conv2d(padding=...)` with even sized kernels. Could you change the tests to reflect this? That means that we should use `nn.Conv2d(kernel_size=kernel_size, padding=same_size_padding(kernel_size))` and test if the input and output shape of an arbitrary image is identical.

Maybe we could add an optional parameter `input` to `same_size_padding`. If its given **and** the `kernel_size` is even, we could pad it. If the `kernel_size` is odd it should be returned unaltered.

pmeier:
> However, I see the problem with the optional parameter input that it is not available at the time of initialisation.

Hm, yeah I didn't think of that.

> For this reason I would suggest that this additional padding is done within the created `ConvBlock`

I think we could generalize this a bit more: we could implement a `PaddedConv2d` that also takes `"valid"` (default), `"same"`, and `"full"` as the parameter `padding`. In `forward` we could call `torch.nn.functional.pad` and simply pass this output to `Conv2d`. Thoughts?

pmeier (commented):
I would make this even more general. Is it OK for you if I push my thoughts to this branch?

jbueltemeier:
Yeah, I'm okay with that.

pmeier:
@jbueltemeier I've settled for same size padding only. Please review if you find obvious shortcomings.

pmeier:
This was more work than I expected, but I think its done now. @jbueltemeier please review and merge if you are happy.

I've implemented the `_SameSizeConvNdMixin` which should be used in double inheritance with any `nn.Conv*` module. With it `SameSizeConv2d` and `SameSizeConvTranspose2d` is just an empty class definition.

Furthermore, this removes the now obsolete functions `same_size_padding`, `same_size_output_padding` and adapts all places where they were used.

PR #176, opened 2020-09-04
Add Sanakoyeu discriminator loss

jbueltemeier:
I will add the tests later, after the changes to the loss modules are done.


pmeier:
Please fix the test errors before I review.

pmeier (commented):
Here is my suggestion how to this in a more clean way:

https://gist.github.com/pmeier/f80bee033af80a50307cfc858ca7f3a3

This is completely untested and thus I didn't push it here. Please review that carefully. 

Furthermore, we need better names than `real` and `imag` (yeah I know that one is really bad :wink:) for the two modes of the discriminator operators. Could you review the GAN literature to find canonical names?

jbueltemeier:
GAN literature always refers to `real` data or `fake` data (sometimes also `generated` data). These names are also used in other implementations in [PyTorch](https://pytorch.org/tutorials/beginner/dcgan_faces_tutorial.html) or [TensorFlow](https://towardsdatascience.com/only-numpy-implementing-gan-general-adversarial-networks-and-adam-optimizer-using-numpy-with-2a7e4e032021). Therefore I would prefer this here, and thus change `imag `only to `fake`.  

Alternatively, the first GAN paper by Goodellow (see [Generative Adversarial Nets](https://arxiv.org/pdf/1406.2661.pdf)) also refers to a:

> a discriminative model that learns to determine whether a sample is from the model distribution or the data distribution. 

Therefore an alternative possibility would be to introduce a `target_distribution`.

pmeier:
In that case we should keep the `real` and `fake` methods to switch. Internally, I think we should replace `_real` with `_target_distribution` to make it more clear what is happening.

pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -309,3 +310,37 @@ class DiscriminatorMultiLayerEncoder(enc.MultiLayerEncoder):
...
+    return conv(
+        in_channels=in_channels,
+        out_channels=1,
+        kernel_size=kernel_size,
+        stride=1,
+        padding=padding,
+    )
+
+
+def get_prediction_modules() -> Dict[str, nn.Module]:
```
pmeier, line 339: Is this still needed?

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+
+    encoder = enc.SequentialEncoder((nn.Conv2d(3, 3, 1),))
+
+    test_op = TestOperator(encoder)
+
+    for mode in (True, False):
+        comparator = torch.ge if mode else torch.lt
+
+        with subtests.test(mode=mode):
+            test_op.real() if mode else test_op.fake()
```
pmeier, line 33:
```suggestion
            test_op.real(mode)
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+    encoder = enc.SequentialEncoder((nn.Conv2d(3, 3, 1),))
+    predictor = nn.Conv2d(3, 3, 1)
+
+    op = paper.PredictionOperator(encoder, predictor)
+
+    for mode in (True, False):
+        comparator = torch.ge if mode else torch.lt
+
+        with subtests.test(mode=mode):
+            op.real() if mode else op.fake()
```
pmeier, line 52:
```suggestion
            op.real(mode)
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+
+    for mode in (True, False):
+        multi_layer_prediction_op.real() if mode else multi_layer_prediction_op.fake()
+        _ = multi_layer_prediction_op(input_image)
+        with subtests.test(mode=mode):
+            with subtests.test("operator mode"):
+                for op in multi_layer_prediction_op.discriminator_operators():
+                    assert op._target_distribution == mode
+
+            with subtests.test("accuracy"):
```
pmeier, line 102: You should almost never test for private attributes: since they might change without warning this test might also break.

```suggestion
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+import pystiche_papers.sanakoyeu_et_al_2018 as paper
+from pystiche import enc, misc
+
+
+def test_EncodingDiscriminatorOperator_call(subtests, input_image):
+    class TestOperator(paper.EncodingDiscriminatorOperator):
+        def input_enc_to_repr(self, image):
+            return image * 2.0
+
+        def calculate_score(self, input_repr):
```
pmeier, line 17: Wouldn't it be better to define this as a function before the class definition and only use it from this method? This way you could reuse the function below.

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+            self.accuracy = self.calculate_accuracy(input_repr)
+            return torch.mean(input_repr)
+
+        def calculate_accuracy(self, input_repr: torch.Tensor) -> torch.Tensor:
```
pmeier, line 21: Same as above.

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -0,0 +1,152 @@
...
+        with subtests.test("layer_weights"):
+            desired_layer_weights = (1e0,) * len(desired_layers)
+            assert layer_weights == pytest.approx(desired_layer_weights)
+
+        with subtests.test("score_weight"):
+            score_weight = 1e0 if impl_params else 1e-3
+            assert prediction_loss.score_weight == pytest.approx(score_weight)
+
+
+def test_discriminator_loss():
```
pmeier, line 150:
```suggestion
def test_discriminator_loss_smoke():
```

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+        pass
+
+
+class PredictionOperator(EncodingDiscriminatorOperator):
+    r"""The prediction loss is a discriminator loss based on the prediction.
+
+    It measures the cross-entropy loss between true labels and predicted labels. The
+    true labels are set depending on the currently set discriminator mode, which can be
+    either ``real`` or ``fake``. In addition, it is calculated how accurate the operator
+    was on the last input.
```
pmeier, line 66: This should be the description of the `EncodingDiscriminatorOperator`, shouldn't it?

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+
+    @abstractmethod
+    def calculate_accuracy(self, input_repr: torch.Tensor) -> torch.Tensor:
+        pass
+
+
+class PredictionOperator(EncodingDiscriminatorOperator):
+    r"""The prediction loss is a discriminator loss based on the prediction.
+
+    It measures the cross-entropy loss between true labels and predicted labels. The
```
pmeier, line 63: This needs to include that an auxiliary classifier is used rather than the raw output of the discriminator.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+        return torch.mean(accuracies)
+
+
+def prediction_loss(
+    impl_params: bool = True,
+    discriminator_multi_layer_encoder: Optional[MultiLayerEncoder] = None,
+    scale_weights: Union[str, Sequence[float]] = "sum",
+    score_weight: Optional[float] = None,
+) -> MultiLayerPredictionOperator:
+    r"""Discriminator loss from :cite:`SKL+2018`.
```
pmeier, line 139: Nope, the `discriminator_loss` is defined below.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+
+    Capture image details at different scales with the ``predictors`` and sum up
+    all losses and accuracies on the different ``scales``.
+
+    Args:
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
+        discriminator_multi_layer_encoder: Trainable
```
pmeier, line 148: Every `MultiLayerEncoder` is trainable by default. The difference in this paper is that it doesn't have to be pre-trained, but will be trained. Still this is detail, that should not make it into this docstring.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+def prediction_loss(
+    impl_params: bool = True,
+    discriminator_multi_layer_encoder: Optional[MultiLayerEncoder] = None,
+    scale_weights: Union[str, Sequence[float]] = "sum",
+    score_weight: Optional[float] = None,
+) -> MultiLayerPredictionOperator:
+    r"""Discriminator loss from :cite:`SKL+2018`.
+
+    Capture image details at different scales with the ``predictors`` and sum up
+    all losses and accuracies on the different ``scales``.
```
pmeier, line 142: Neither `predictors` nor `scales` are defined in this docstring.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,255 @@
...
+        return self.real(False)
+
+    def get_accuracy(self) -> torch.Tensor:
+        accuracies = torch.stack([op.accuracy for op in self.discriminator_operators()])
+        return torch.mean(accuracies)
+
+
+def prediction_loss(
+    impl_params: bool = True,
+    discriminator_multi_layer_encoder: Optional[MultiLayerEncoder] = None,
```
pmeier, line 135: Maybe `multi_layer_discriminator`? If we change this, we should also change `DiscriminatorMultiLayerEncoder` to `MultiLayerDiscriminator`.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
jbueltemeier: I just agreed too quickly. Isn't it better to simply call this `multi_layer_encoder`, as this is what is needed here.

pmeier: True. What do you think of the change to `DiscriminatorMultiLayerEncoder`?

jbueltemeier: I would think that the previous name should remain, as this is only a special structure of a `MultiLayerEncoder`. Therefore, the naming should remain the same as for similar classes such as `VGGMultiLayerEncoder`.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+class EncodingDiscriminatorOperator(ops.EncodingRegularizationOperator):
+    r"""Abstract base class for all discriminator operators working in an encoded space.
+
+    In this operator a discriminator mode can be set, in whose dependence the output can
+    be influenced. This can be either ``real`` or ``fake``. In addition, it is
+    calculated how accurate the operator was on the last input.
+
+    Args:
+        encoder: Encoder that is used to encode the target and input images.
+        score_weight: Score weight of the operator. Defaults to ``1.0``.
```
pmeier, line 38:
```suggestion
        score_weight: Score weight of the operator. Defaults to ``1.0``.
        
    Attributes:
        accuracy: ADDME
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+    "transformed_image_loss",
+]
+
+
+class EncodingDiscriminatorOperator(ops.EncodingRegularizationOperator):
+    r"""Abstract base class for all discriminator operators working in an encoded space.
+
+    In this operator a discriminator mode can be set, in whose dependence the output can
+    be influenced. This can be either ``real`` or ``fake``. In addition, it is
+    calculated how accurate the operator was on the last input.
```
pmeier, line 34: Please move this description into the docstrings of `.real()` and `.fake()`

```suggestion
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+
+
+class PredictionOperator(EncodingDiscriminatorOperator):
+    r"""The prediction loss is a discriminator loss based on the prediction.
+
+    The prediction consists of the output of an ``predictor`` which processes the
+    output of the ``encoder``.
+
+    It measures the cross-entropy loss between true labels and predicted labels. The
+    true labels are set depending on the currently set discriminator mode.
```
pmeier, line 71: This should be clear from the docstring of the `EncodingDiscriminatorOperator` or any of its methods. If it is not, please fix it there.

```suggestion
    It measures the cross-entropy loss between true labels and predicted labels. 
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+
+    Args:
+        encoder: Encoder that is used to encode the target and input images.
```
pmeier, line 74: What is the difference between target and input images here?

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
pmeier, line 71: If I didn't know the structure, I would be confused by this. 

```suggestion
    r"""Partial discriminator loss based on the :class:`torch.nn.BCELoss`.

    The class prediction is generated by the ``predictor``, which acts an 
    auxiliary classifier.
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+        predictor: Auxiliary classifier used to process the encodings into a prediction.
```
pmeier, line 75:
```suggestion
        predictor: Auxiliary classifier used to predict real or fake from the encodings of the ``encoder``.
```

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+        return torch.mean(accuracies)
+
+
+def prediction_loss(
+    impl_params: bool = True,
+    multi_layer_encoder: Optional[MultiLayerEncoder] = None,
+    scale_weights: Union[str, Sequence[float]] = "sum",
+    score_weight: Optional[float] = None,
+) -> MultiLayerPredictionOperator:
+    r"""Prediction loss indicates whether the input is real or fake.
```
pmeier, line 144: That is not true. A high loss simply indicates a bad classification. That might be a real image confidently classified as fake or vice versa. From a high (or with reversed logic from a low) loss you can't tell whether the input was real or fake.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,261 @@
...
+
+    Capture image details at different scales with an auxiliary classifier and sum up
+    all losses and accuracies on different layers of the
+    :class:`~pystiche.enc.MultiLayerEncoder`.
```
pmeier, line 148: This should make clear that this is the partial discriminator loss for a single image.

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -1,7 +1,184 @@
...
+    layers = [str(index) for index in range(3)]
+    modules = [(layer, nn.Conv2d(3, 3, 1)) for layer in layers]
+    multi_layer_encoder = enc.MultiLayerEncoder(modules)
+
+    multi_layer_prediction_op = paper.MultiLayerPredictionOperator(
+        multi_layer_encoder, layers, get_encoding_op
+    )
+
+    for mode in (True, False):
+        multi_layer_prediction_op.real() if mode else multi_layer_prediction_op.fake()
```
pmeier, line 87:
```suggestion
        multi_layer_prediction_op.real(mode)
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -1,7 +1,184 @@
...
+        _ = multi_layer_prediction_op(input_image)
```
pmeier, line 88:
```suggestion
        multi_layer_prediction_op(input_image)
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -1,7 +1,184 @@
...
+        with subtests.test(mode=mode):
+            desired = torch.mean(
+                torch.stack(
+                    [
+                        op.accuracy
+                        for op in multi_layer_prediction_op.discriminator_operators()
+                    ]
+                )
+            )
```
pmeier, line 97: Please split this up to make it more readable:

```python
accuracies = ...
desired = torch.mean(torch.stack(accuracies))
```

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -1,7 +1,184 @@
...
+            score_weight = 1e0 if impl_params else 1e-3
+            assert prediction_loss.score_weight == pytest.approx(score_weight)
+
+
+def attach_method_mock(mock, method, **attrs):
+    if "name" not in attrs:
+        attrs["name"] = f"{mock.name}.{method}()"
+
+    method_mock = unittest.mock.Mock(**attrs)
+    mock.attach_mock(method_mock, method)
```
pmeier, line 142: If I'm not mistaken, this is almost an exact copy of `tests.mocks.attach_method_mock`. Why not use this one?

tests/unit/sanakoyeu_et_al_2018/test_loss.py
```diff
@@ -1,7 +1,184 @@
...
+
+
+@pytest.fixture
+def prediction_loss_mocks(mocker):
+    attrs = {}
+    attrs["side_effect"] = lambda image: pystiche.LossDict([("0", torch.mean(image))])
+
+    mock = mocker.Mock(**attrs)
```
pmeier, line 150: Why this weird initalization?

```suggestion
    mock = mocker.Mock(side_effect=lambda image: pystiche.LossDict([("0", torch.mean(image))]))
```

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -1,18 +1,271 @@
...
+
+    def real(self, mode: bool = True) -> "MultiLayerPredictionOperator":
+        for op in self.discriminator_operators():
+            op.real(mode)
+        return self
+
+    def fake(self) -> "MultiLayerPredictionOperator":
+        return self.real(False)
+
+    def get_accuracy(self) -> torch.Tensor:
```
pmeier, line 143: Please add a docstring.

pmeier (changes requested):

pmeier (approved):

PR #178, opened 2020-09-08
fix same size padding

pmeier:
apply the changes from #175


pystiche_papers/sanakoyeu_et_al_2018/_modules.py
```diff
@@ -48,12 +37,22 @@ def conv(
     in_channels: int,
     out_channels: int,
     kernel_size: Union[Tuple[int, int], int],
-    stride: Union[Tuple[int, int], int] = 2,
```
pmeier, line 51: @jbueltemeier I don't know why this was `2` by default? Was there a reason for this? Tests are passing so it seems this is irrelevant for all other internal functions.

jbueltemeier: No, there was no particular reason.

jbueltemeier (approved):

PR #179, opened 2020-09-08
Auto pad for pooling

pmeier:
This adds same size pooling similar to #175, since this is needed for https://github.com/pmeier/pystiche_papers/pull/174#issuecomment-688774781


pmeier:
This also renames `SameSize*` to `AutoPad*` since for `stride > 1` the output size is not the same size as the input but rather a multiple of `stride`.

PR #177, opened 2020-09-08
Add data functionalities to sanakoyeu_et_al_2018

pmeier:
Supersedes #162. @jbueltemeier merge if you are happy.

PR #183, opened 2020-09-10
add data augmentation to sanakoyeu_et_al_2018

pmeier:
This adds the image augmentation pipeline that was split off from #162 (and later #177 ). For the augmentation transforms we pull `kornia` as dependency. Given that `pystiche` has already started to replace the custom transforms with `kornia` and it will be a dependency in the future anyway, this should be fine. 

ToDo:

- [x] Tests

### DIFF
--- a/pystiche_papers/sanakoyeu_et_al_2018/_augmentation.py
+++ b/pystiche_papers/sanakoyeu_et_al_2018/_augmentation.py
@@ -1,0 +1,469 @@
+from typing import Any, Dict, List, Tuple, Union, cast
+
+import kornia
+import kornia.augmentation.functional as F
+from kornia.augmentation.random_generator import _adapted_uniform, random_prob_generator
+
+import torch
+from torch import nn
+from torch.nn.functional import pad
+
+import pystiche
+from pystiche.image import extract_image_size
+from pystiche.misc import to_2d_arg
+
+__all__ = ["augmentation"]
+
+
+class AugmentationBase(pystiche.ComplexObject, kornia.augmentation.AugmentationBase):
+    def _properties(self) -> Dict[str, Any]:
+        dct = super()._properties()
+        if self.return_transform:
+            dct["return_transform"] = True
+        return dct
+
+
+class RandomRescale(AugmentationBase):
+    def __init__(
+        self,
+        factor: Union[Tuple[float, float], float],
+        probability: float = 50e-2,
+        interpolation: str = "bilinear",
+        align_corners: bool = False,
+    ):
+        super().__init__(return_transform=False)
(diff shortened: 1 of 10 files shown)


pystiche_papers/sanakoyeu_et_al_2018/_augmentation.py
```diff
@@ -0,0 +1,492 @@
...
+    high: Union[float, torch.Tensor],
+    same_on_batch: bool = False,
+) -> torch.Tensor:
+    return _adapted_uniform(shape, low, high + 1 - 1e-6, same_on_batch).int()
+
+
+class RandomCrop(AugmentationBase):
+    def __init__(
+        self,
+        size: Union[Tuple[int, int], int],
```
jbueltemeier, line 217: The authors use a default `crop_size` of `(256,256)`. See [here](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/img_augm.py#L26).

pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -84,22 +85,25 @@ def _properties(self) -> Dict[str, Any]:
...
-) -> transforms.ComposedTransform:
-    transform = style_image_transform(edge_size)
+    impl_params: bool = True, edge_size: int = 768, train: bool = False,
+) -> nn.Sequential:
+    transform = style_image_transform(edge_size=edge_size, train=train)
     if not impl_params:
         return transform
 
-    return transforms.ComposedTransform(transforms.Rescale(2.0), transform)
+    return nn.Sequential(transforms.Rescale(2.0), *transform.children())
```
jbueltemeier, line 106: During `stylization`, only the smaller side of the image is resized to the `image_size`. See [here](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L492-L494).

pystiche_papers/sanakoyeu_et_al_2018/_augmentation.py
pmeier: That is true, but they are [not using the default](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L271) during training.

pystiche_papers/sanakoyeu_et_al_2018/_data.py
pmeier: This refers to https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/prepare_dataset.py#L133. We should have done this in #177. I'll send a follow-up PR.

jbueltemeier (approved):

PR #187, opened 2020-09-10
Test downloads on a fixed schedule

### LINKED ISSUE #186: Test downloadability less often
Currently, we are testing the downloadability of images, datasets, models, ... within the regular test suite for every PR. This is not only slow, but also unnecessary, since it might generate a lot of traffics on days with a lot of commits. On the other hand, if no commit is performed within a longer interval, breaking downloads might not be detected at all.

To overcome this I propose we run the download tests on a fixed schedule.

pmeier:
Closes #186. During the normal test suite, the download tests will only run if a download test or the hub or test config  changed within a PR. As opposed by a PR trigger, the scheduled run also tests large downloads. A GitHub Actions bot will create an issue if a scheduled run fails.


tests/unit/johnson_alahi_li_2016/test_modules.py
```diff
@@ -308,22 +281,46 @@ def test_transformer_smoke(subtests, image_medium):
         assert image_medium.size() == output_image.size()
 
 
-@pytest.mark.large_download
```
pmeier, line 311: The download was performed twice before: here and within the hub tests. Now we only check if `select_url` is called correctly and the correct `state_dict` is loaded here. The download stays the same in the hub tests.

PR #189, opened 2020-09-11
Add Sanakoyeu 2018 transformer loss

jbueltemeier:
Rebase this after [#174](https://github.com/pmeier/pystiche_papers/pull/174) and [#176](https://github.com/pmeier/pystiche_papers/pull/176) are merged.

### DIFF
--- a/pystiche_papers/sanakoyeu_et_al_2018/_loss.py
+++ b/pystiche_papers/sanakoyeu_et_al_2018/_loss.py
@@ -272,7 +272,7 @@ def transformed_image_loss(
     impl_params: bool = True,
     score_weight: Optional[float] = None,
 ) -> ops.FeatureReconstructionOperator:
-    r"""Transformed_image_loss from from :cite:`SKL+2018`.
+    r"""Transformed_image_loss from :cite:`SKL+2018`.
 
     Args:
         transformer_block::class:`~pystiche_papers.sanakoyeu_et_al_2018.TransformerBlock`
@@ -351,7 +351,7 @@ def transformer_loss(
     transformed_image_operator: Optional[ops.FeatureReconstructionOperator] = None,
     style_loss: Optional[MultiLayerPredictionOperator] = None,
 ) -> PerceptualLoss:
-    r"""Transformer_loss from from :cite:`SKL+2018`.
+    r"""Transformer_loss from :cite:`SKL+2018`.
 
     Args:
         encoder: :class:`~pystiche.enc.SequentialEncoder`.
--- a/pystiche_papers/sanakoyeu_et_al_2018/_loss.py
+++ b/pystiche_papers/sanakoyeu_et_al_2018/_loss.py
@@ -1,14 +1,13 @@
 from abc import abstractmethod
 from collections import OrderedDict
-from typing import Iterator, Optional, Sequence, Union, cast
+from typing import Any, Dict, Iterator, Optional, Sequence, Union, cast
 
 import torch
 import torch.nn as nn
 from torch.nn.functional import binary_cross_entropy_with_logits
 
-from pystiche import enc, ops
+from pystiche import enc, loss, ops
 from pystiche.enc import Encoder, MultiLayerEncoder, SequentialEncoder
-from pystiche.loss.perceptual import PerceptualLoss
 
 from ._modules import (
     DiscriminatorMultiLayerEncoder,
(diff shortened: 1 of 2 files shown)


pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -289,3 +293,94 @@ def transformed_image_loss(
...
+
+class FeatureReconstructionOperator(ops.FeatureReconstructionOperator):
+    def __init__(
+        self,
+        encoder: SequentialEncoder,
+        score_weight: float = 1.0,
+        impl_params: bool = True,
+    ):
+        super().__init__(encoder, score_weight=score_weight)
+        self.impl_params = impl_params
```
pmeier, line 306: Never store `impl_params`, but only store the changes. For example here it could be `reduction = "mae" if impl_params else "mse"` or something similar. Don't do this though and look at the next comment first.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -289,3 +293,94 @@ def transformed_image_loss(
...
+
+    def calculate_score(  # type: ignore[override]
+        self, input_repr: torch.Tensor, target_repr: torch.Tensor, ctx: None
+    ) -> torch.Tensor:
+        if self.impl_params:
+            # https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L194
+            # https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/module.py#L177-L178
+            return torch.mean(torch.abs(input_repr - target_repr))
```
pmeier, line 314: Changing the reduction is not a change that should happen here since it is a core concept of the `ops.FeatureReconstructionOperator`. I suggest you implement an `MAEReconstructionOperator` and return this or the other based on `impl_params` from `style_aware_content_loss`.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -289,3 +293,94 @@ def transformed_image_loss(
...
+        encoder, score_weight=score_weight, impl_params=impl_params
+    )
+
+
+def transformer_loss(
+    encoder: SequentialEncoder,
+    impl_params: bool = True,
+    style_aware_content_operator: Optional[ops.FeatureReconstructionOperator] = None,
+    transformed_image_operator: Optional[ops.FeatureReconstructionOperator] = None,
+    style_loss: Optional[MultiLayerPredictionOperator] = None,
```
pmeier, line 352: Why would ever pass these instead using the default?

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -8,6 +8,7 @@
 
 from pystiche import enc, ops
 from pystiche.enc import Encoder, MultiLayerEncoder, SequentialEncoder
+from pystiche.loss.perceptual import PerceptualLoss
```
pmeier, line 11: You could just use `from pystiche import loss` and use it like that.

pmeier (changes requested):

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -289,3 +292,122 @@ def transformed_image_loss(
...
+    def target_enc_to_repr(self, enc: torch.Tensor) -> Tuple[torch.Tensor, None]:
+        return self.enc_to_repr(enc), None
+
+    def calculate_score(
+        self,
+        input_repr: torch.Tensor,
+        target_repr: torch.Tensor,
+        ctx: Optional[torch.Tensor],
+    ) -> torch.Tensor:
+        return torch.mean(torch.abs(input_repr - target_repr))
```
pmeier, line 331: Although the implementation is simple, we should use `torch`s [builtin implementation](https://pytorch.org/docs/stable/nn.functional.html#torch.nn.functional.l1_loss) here.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -289,3 +292,122 @@ def transformed_image_loss(
...
+        style_aware_content_kwargs: Optional parameters for the
+            :func:`style_aware_content_loss`.
+        transformed_image_kwargs: Optional parameters for the
+            :func:`transformed_image_loss`.
+        style_loss_kwargs: Optional parameters for the :func:`prediction_loss`.
+
+    """
+    if style_aware_content_kwargs is None:
+        style_aware_content_kwargs = {}
+
```
pmeier, line 389: Nit

```suggestion
```

pmeier (approved):
Two minor comments. Afterwards, this is good to merge.

PR #190, opened 2020-09-11
Fix padding docs in sanakoyeu 2018

jbueltemeier:
Due to the use of [#178](https://github.com/pmeier/pystiche_papers/pull/178) some changes to the docs became necessary.


pmeier (approved):
Thanks, I missed that in #182.

PR #201, opened 2020-09-17
fix sanakoyeu_et_al_2018 image transforms

pmeier:
This fixes multiple issues with respect to the image transformations and augmentation:

- Currently, the augmentation defines a separate transform including a random crop after the image transform (`resize -> crop -> augmentation`). Thus, the random crop is performed twice. This PR splits the augmentation in two parts to account for that (`resize -> pre-crop augmentation -> crop -> post-crop-augmentation`)
- Currently, the content image transform includes an prefixed upscaling of factor 2. This is better placed at the content dataset.

### DIFF
--- a/pystiche_papers/sanakoyeu_et_al_2018/_data.py
+++ b/pystiche_papers/sanakoyeu_et_al_2018/_data.py
@@ -17,8 +17,9 @@ from pystiche.image.utils import extract_edge_size, extract_image_size
 from pystiche.misc import to_2d_arg, verify_str_arg
 
 from ._augmentation import (
-    AugmentationBase,
+    AugmentationBase2d,
     _adapted_uniform,
+    generate_vertices_from_size,
     post_crop_augmentation,
     pre_crop_augmentation,
 )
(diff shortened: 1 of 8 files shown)


pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -85,29 +93,171 @@ def _properties(self) -> Dict[str, Any]:
...
+        return dct
+
+
+def image_transform(
+    impl_params: bool = True, edge_size: int = 768, train: bool = False
 ) -> nn.Sequential:
     transforms_: List[nn.Module] = [
-        ClampSize(),
-        transforms.ValidRandomCrop(image_size),
-        OptionalGrayscaleToFakegrayscale(),
```
pmeier, line 96: @jbueltemeier Do you remember why we included this? I can't find any evidence of this in their implementation? If not, I think it is OK to remove it.

jbueltemeier: This was only introduced to get the `image_size `of 768x768, because there was no `augmentation `before. So you are right that it is no longer needed and can be removed.

jbueltemeier: Ok, I just don't understand why it makes sense here to split the augmentation. 

What we need is if:

1. `impl_params=True and train=True`: a sequence a sequence of `ClampSize -> augmentation`  with internal cropping

2. `impl_params=True and train=False`: Resize the smallest side of the image to the image_size and optionally a transformation back to the original image size. ([resize](https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L494), [optional back](https://github.com/pmeier/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/model.py#L506)).

3.  `impl_params=false`: An interpretation of how to handle images that are smaller than the `image_size` in the quote below. So maybe a sequence of `OptionalUpsample -> ValidRandomCrop`.
> trained from scratch on randomly cropped image patches of size 768×768 pix.

pmeier: In the reference implementation the training pipeline goes like this:

1. `ClampSize`
2. `RandomRescale`
3. `DynamicSizePad2d`
    1. `RandomRotation`
    2. `RandomAffine`
4. `RandomCrop`
5. `RandomHSVJitter`
6. `RandomHorizontalFlip`

I've aggregated 2. and 3. to `pre_crop_augmentation` and 5. and 6. to `post_crop_augmentation`.

That leaves us with the following cases

- `impl_params and train`: see pipeline above
- `impl_params and not train`: remove `pre_crop_augmentation` and `post_crop_augmentation` leaving only `ClampSize` and `RandomCrop`
- `not impl_params`: I think (and if I understand you correctly you agree) the transformation that achieves 
    > randomly cropped image patches of size 768×768 pix

    is an optional upsample to 768 pixels on the shorter edge and subsequently a valid random crop.

As is, my implementation does exactly that.

---

> Resize the smallest side of the image to the image_size and optionally a transformation back to the original image size.

This is only for the images that you want to stylize after the training. There is no reason to pass them through these transforms at all. If we want this functionality at all, and I'm not sure about that yet, it should be implemented in `paper.nst()`.

jbueltemeier: Okay, I agree with you on cases one and three. Only the case `impl_params and not train` I do not find in the implementation. This should be the case `is_inference` in their implementation and there only a `resize` is done instead of the `crop`. See [here](https://github.com/CompVis/adaptive-style-transfer/blob/07a3b3fcb2eeed2bf9a22a9de59c0aea7de44181/img_augm.py#L60).  Or am I still thinking wrong?

pmeier: No, you were right. `train=True` should always be the case, since `image_transforms` are only used during training. I've removed it in the recent commits.

PR #209, opened 2020-11-02
Add Sanakoyeu_et_al_2018 nst training 

jbueltemeier:


pystiche_papers/sanakoyeu_et_al_2018/_loss.py
```diff
@@ -178,7 +179,6 @@ def transformer_loss(
     impl_params: bool = True,
     style_aware_content_kwargs: Optional[Dict[str, Any]] = None,
     transformed_image_kwargs: Optional[Dict[str, Any]] = None,
-    style_loss_kwargs: Optional[Dict[str, Any]] = None,
```
pmeier, line 181: Why did you remove this?

pystiche_papers/sanakoyeu_et_al_2018/_nst.py
```diff
@@ -210,3 +222,114 @@ def optim_loop(transformer: nn.Module) -> nn.Module:
             transformer_lr_scheduler.step(epoch)
 
     return transformer
+
+
+def training(
+    content_image_loader: DataLoader,
+    style_image_loader: DataLoader,
```
pmeier, line 229: Couldn't we also accept a `str` and use this as root to load the content and style dataset?

```suggestion
    content_image_loader: Union[DataLoader, str],
    style_image_loader: Union[DataLoader, str],
```

In case of the style dataset we could also introduce a `style: Optional[str] = None` parameter this is passed through the dataset. If this `is None` and `isinstance(style_image_loader, str)` we should raise an `ValueError` that `style` is needed for the style dataset. If on the other hand, `isinstance(style_image_loader, DataLoader)` we can simply ignore the `style` parameter or emit a warning that it has no effect.

pystiche_papers/sanakoyeu_et_al_2018/_loss.py
jbueltemeier: I have removed this one because the `prediction_loss`  is not to be changed anymore. The `MultiPredictionOperator` (`prediction_loss`) should be the same for the `transformer_loss` and `discriminator_loss`.

pystiche_papers/sanakoyeu_et_al_2018/_nst.py
```diff
@@ -210,3 +223,137 @@ def optim_loop(transformer: nn.Module) -> nn.Module:
...
+) -> nn.Module:
+    r"""Training a transformer for the NST.
+
+    Args:
+        content_image_loader: Content images used as input for the ``transformer`` or
+            string to use a predefined image_loader.
+        style_image_loader: Style images used as input for the ``discriminator`` or
+            string to use a predefined image_loader.
+        style: Optional style string which is needed if the 'style_image_loader' is a
+            string.
```
pmeier, line 242:
```suggestion
        style: Style to train the transformer on. See :func:`style_dataset` for details.
            .. note::
            
              This parameter has no effect if a :class:`~torch.utils.data.DataLoader` is provided for ``style_image_loader``
```

pystiche_papers/sanakoyeu_et_al_2018/_nst.py
```diff
@@ -210,3 +223,137 @@ def optim_loop(transformer: nn.Module) -> nn.Module:
...
+        impl_params: If ``True``, uses the parameters used in the reference
+            implementation of the original authors rather than what is described in
+            the paper.
+
```
pmeier, line 246:
```suggestion

Raises:
    ValueError: If no ``style`` is passed, but ``style_image_loader`` is passed as dataset root.
```

pystiche_papers/sanakoyeu_et_al_2018/_nst.py
```diff
@@ -210,3 +223,137 @@ def optim_loop(transformer: nn.Module) -> nn.Module:
...
+            root = style_image_loader
+            dataset = style_dataset(root, style=style, impl_params=impl_params)
+            style_image_loader = image_loader(dataset, impl_params=impl_params)
+        else:
+            raise ValueError(
+                "The style is currently '"
+                + str(style)
+                + "', please use a valid string for the style because it is needed for "
+                + "the style_dataset."
+            )
```
pmeier, line 266: 1. You should use `f-strings` since they are more readable (and faster, but that is not an issue here). This would also remove the need to concatenate the strings with addition.
2. The message is to explicit. 

```suggestion
            raise ValueError(
                "The parameter 'style' cannot be omitted if 'style_image_dataset' is passed as dataset root."
            )
```

pmeier (changes requested):
As a rule of thumb: document for intent and not for content. For example:

- the type of the `content_image_loader` is not as important as documentation that it will be used as the root of the dataset if it is not a DataLoader. The type will be automatically added from the type hints.
- The actual value of the `style` is not important but rather if it is passed or not. Since we cannot perform a check if the style is valid here, there is no use in including it into the error message.

pmeier (approved):

PR #210, opened 2020-11-06
Split sanakoyeu_et_al_2018 transformer and discriminator modules

jbueltemeier:
Currently all modules of the `transformer` and the `discriminator` are stored in two file (modules.py and loss.py), which makes it very confusing to divide between the modules of the `transformer` and the `discriminator`. Therefore, here is a suggestion for a better subdivision to make the implementation clearer and thus more comprehensible.


pmeier (commented):
This is a good start. If I see this right, `_loss.py` and `_modules.py` now only contain things that used in the transformer as well as the discriminator, correct? It might be confusing to have a `_loss.py`, but actually define all the high-level losses somewhere else. Do you have a idea how to fix that? Maybe move the common thing to `_utils.py` or `_common.py`?

jbueltemeier:
>  If I see this right, `_loss.py` and `_modules.py` now only contain things that used in the transformer as well as the discriminator, correct?

1. That's not entirely true. `_modules.py` contains only modules that occur in both or neither (`TransformerBlock`) of them.
In `_loss.py`  all high-level losses are defined, which will be used as `criterion` for the `transformer` or the `discriminator`, just like in the other implementations.

> It might be confusing to have a _loss.py, but actually define all the high-level losses somewhere else.

2. All high-level losses are defined there. `_loss.py` contains only the `DiscriminatorLoss` as criterion for the `discriminator`, as well as the `PerceptualLoss` (`transformer_loss`) and the corresponding losses `style_aware_content_loss` and `tranformed_image_loss`.

> Maybe move the common thing to _utils.py or _common.py?

3.  `_modules.py` contain all common components, but also the TransformerBlock, which is not part of the Transformer, but is only used for the transformation of the input images and the output images of the Transformer for another `loss`. So in my opinion it would not make sense to pack these components in `_utils.py` or to open `_common.py`.

pmeier (approved):
That makes sense, thanks for the explanation!

PR #211, opened 2020-11-11
fix li_wand_2016 target_transforms

pmeier:
With `impl_params=True` we cannot use the default `pystiche.ops.MRFOperator.scale_and_rotate_transforms`, but need to use custom code instead. With this the replication now yields almost the same results as the reference implementation:

# Theirs

res_3_100

# Our

fig_6__top

Todo:

- [x] tests
- [x] documentation

### DIFF
--- a/replication/li_wand_2016/main.py
+++ b/replication/li_wand_2016/main.py
@@ -28,7 +28,7 @@ def figure_6(args):
             f"Replicating the {position} half of figure 6 " f"with {params} parameters"
         )
 
-        # https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L67
+        # https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L66
         style_loss_kwargs = {
             "patch_size": 2,
             "stride": 1,
--- a/replication/li_wand_2016/main.py
+++ b/replication/li_wand_2016/main.py
@@ -5,7 +5,6 @@ from os import path
 import pystiche_papers.li_wand_2016 as paper
 from pystiche.image import write_image
 from pystiche.misc import get_device
-from pystiche.ops import MRFOperator
 from pystiche.optim import OptimLogger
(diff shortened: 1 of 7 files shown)


replication/li_wand_2016/main.py
```diff
@@ -29,14 +28,13 @@ def figure_6(args):
...
-        # https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L65-L70
+        # https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L67
         style_loss_kwargs = {
-            "target_transforms": MRFOperator.scale_and_rotate_transforms(
-                num_scale_steps=1,
-                scale_step_width=5e-2,
-                num_rotate_steps=1,
-                rotate_step_width=7.5,
-            )
+            "patch_size": 2,
```
jbueltemeier, line 33: Should this not be `patch_size=3`? This should correspond to the 9 entry in line [L67](https://github.com/pmeier/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/run_trans.lua#L67).

replication/li_wand_2016/main.py
```diff
@@ -29,14 +28,13 @@ def figure_6(args):
...
+            "stride": 1,
```
jbueltemeier, line 34: Same as above. This should be `stride=2`.

jbueltemeier (commented):
You are right with this transformation the images look very similar, only the third eye on the chin is still missing. However, I have the assumption that my comments will eventually fix this.

replication/li_wand_2016/main.py
pmeier: Two things:

1. My comment referenced the wrong line. [L66]() is the relevant one.
2. Regardless of that, my local copy had these changes and I just used this. Good catch!

replication/li_wand_2016/main.py
pmeier: Same as above.

pmeier:
With the corrected parameters we get the following results (theirs / ours)

res_3_100 fig_6__top

jbueltemeier (approved):

PR #215, opened 2020-11-12
Add content images from the repository

jbueltemeier:
Add some of the content images used to compare the replication with the original implementation.

These images are used in the implementation when comparing the functionality for different styles. See [here](https://github.com/CompVis/adaptive-style-transfer/tree/gh-pages/images/experiments/comparison_with_others).

### DIFF
--- a/pystiche_papers/sanakoyeu_et_al_2018/_data.py
+++ b/pystiche_papers/sanakoyeu_et_al_2018/_data.py
@@ -2,6 +2,7 @@ import os
 import re
 from os import path
 from typing import Any, Dict, List, Optional, Sized, Tuple, Union, cast
+from urllib.parse import urljoin
 
 import kornia
 from kornia.augmentation.functional import apply_crop, compute_crop_transformation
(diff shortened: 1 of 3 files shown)


pystiche_papers/sanakoyeu_et_al_2018/_data.py
```diff
@@ -247,6 +253,62 @@ def image_transform(impl_params: bool = True, edge_size: int = 768) -> nn.Sequen
     return nn.Sequential(*transforms_)
 
 
+def images() -> DownloadableImageCollection:
+    base_sanakoyeu = "https://raw.githubusercontent.com/CompVis/adaptive-style-transfer/gh-pages/images/experiments/comparison_with_others/"
```
pmeier, line 257: I suggest you implement something like

```python
def url(
    id: int,
    file: str,
    base: str = "https://raw.githubusercontent.com/CompVis/adaptive-style-transfer/gh-pages/images/experiments/comparison_with_others/",
) -> str:
    return urljoin(base, f"{id:04d}/{file}")
```

and use that throughout this function. The double `urljoin` feels clumsy.

pmeier (commented):
Could you also add this paper to 

https://github.com/pmeier/pystiche_papers/blob/9f2acb82191353c55370326a5f4a91cd7528d969/tests/download/test_images.py#L12-L18

to enable download tests for these images?

pmeier (approved):

Issue #219, opened 2020-11-13
GramOperator anomaly for ulyanov_et_al_2016


pmeier:
Since the initial PR (#17) the following two lines are included in the constructor of `pystiche.ulyanov_et_al_2016.GramOperator`:

https://github.com/pmeier/pystiche_papers/blob/9f2acb82191353c55370326a5f4a91cd7528d969/pystiche_papers/ulyanov_et_al_2016/_loss.py#L94-L95

While the latter is simply ignored throughout the operator, the former changes the computation of the representation:

https://github.com/pmeier/pystiche_papers/blob/9f2acb82191353c55370326a5f4a91cd7528d969/pystiche_papers/ulyanov_et_al_2016/_loss.py#L103-L109

Both parameters did not get a reference link in #109. @jbueltemeier could you have a look?

jbueltemeier:
Fixed in [#221](https://github.com/pmeier/pystiche_papers/pull/221).

PR #221, opened 2020-11-16
Fix normalization for GramOperator from ulyanov_et_al_2016

jbueltemeier:
Noted in [#219](https://github.com/pmeier/pystiche_papers/issues/219). I added the permalink and removed the superfluous line.

### DIFF
--- a/pystiche_papers/ulyanov_et_al_2016/_loss.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_loss.py
@@ -91,8 +91,8 @@ class GramOperator(ops.GramOperator):
         self, encoder: enc.Encoder, impl_params: bool = True, **gram_op_kwargs: Any,
     ):
         super().__init__(encoder, **gram_op_kwargs)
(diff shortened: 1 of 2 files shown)


pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -91,8 +91,8 @@ def __init__(
         self, encoder: enc.Encoder, impl_params: bool = True, **gram_op_kwargs: Any,
     ):
         super().__init__(encoder, **gram_op_kwargs)
+        # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L38
```
pmeier, line 94: As far as I see, this is the only normalization that they perform, correct? In that case, we need to create the operator with `normalize=False`. Otherwise we will be normalizing the Gram matrix by the height and width.

pmeier (approved):

PR #248, opened 2020-12-10
Improvement ulyanov_et_al_2016

jbueltemeier:
So far, this implementation has not produced good results. This was because in the reference 
implementation the backward pass of the gradient is normalized and this is missing 
in this implementation. The results are now better, but still with measures, probably due to different hyperparameter that are too high or too low. Therefore, similar results as in the paper can now be created if the `style_weight` is changed.

Therefore in the following two points which should be discussed in advance:

1. Due to the changed hyperparameter setting from [#244](https://github.com/pmeier/pystiche_papers/pull/244) it is now easier to adjust the hyperparameter. Therefore i would adjust the `style_weight` for this replication so that besides the results with the default parameters similar results to the paper are created.

2. One problem with the replication of the master branch is that the images are too small and have to be enlarged for training. I currently added an additional `resize` in the `content_image_transform` before the `ValidRandomCrop`, but this is not 
implemented in the reference implementation.

### DIFF
--- a/pystiche_papers/ulyanov_et_al_2016/_data.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_data.py
@@ -187,9 +187,7 @@ def images() -> DownloadableImageCollection:
             md5="7fdd9603a5182dcef23d7fb1c5217888",
         ),
     }
-    return DownloadableImageCollection(
-        {**content_images, **style_images}
-    )
+    return DownloadableImageCollection({**content_images, **style_images})
 
 
 def dataset(
--- a/pystiche_papers/ulyanov_et_al_2016/_data.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_data.py
@@ -123,15 +123,6 @@ def images() -> DownloadableImageCollection:
             md5="dc9ad203263f34352e18bc29b03e1066",
             file="tuebingen_neckarfront__andreas_praefcke.jpg",
         ),
-        "che_high": DownloadableImage(
-            "https://upload.wikimedia.org/wikipedia/commons/5/58/CheHigh.jpg",
-            title="CheHigh",
-            author="Alberto Korda",
-            date="1960",
-            license=ExpiredCopyrightLicense(1960),
-            md5="62af957a89fdfcb9618570b6d836d054",
-            file="CheHigh.jpg",
-        ),
         "tower_of_babel": DownloadableImage(
             "https://upload.wikimedia.org/wikipedia/commons/f/fc/Pieter_Bruegel_the_Elder_-_The_Tower_of_Babel_%28Vienna%29_-_Google_Art_Project_-_edited.jpg",
             title="The Tower of Babel",
@@ -151,31 +142,6 @@ def images() -> DownloadableImageCollection:
         ),
     }
 
-    texture_base_ulyanov = urljoin(base_ulyanov, "textures/")
-    base_ulyanov_suppl_texture = "https://raw.githubusercontent.com/DmitryUlyanov/texture_nets/texture_nets_v1/supplementary//texture_models/"
-
-    texture_images = {
-        "cezanne": DownloadableImage(
-            urljoin(texture_base_ulyanov, "cezanne.jpg"),
-            md5="fab6d360c361c38c331b3ee5ef0078f5",
-        ),
-        "bricks": DownloadableImage(
-            urljoin(base_ulyanov_suppl_texture, "bricks.png"),
-            md5="1e13818e1fbefbd22f110a1c2f781d40",
-        ),
-        "pebble": DownloadableImage(
-            urljoin(base_ulyanov_suppl_texture, "pebble.png"),
-            md5="5b5e5aa6c579e42e268058a94d683a6c",
-        ),
-        "pixels": DownloadableImage(
-            urljoin(base_ulyanov_suppl_texture, "pixelcity_windows2.jpg"),
-            md5="53026a8411e7c26e959e36d3223f3b8f",
-        ),
-        "peppers": DownloadableImage(
-            urljoin(base_ulyanov_suppl_texture, "red-peppers256.o.jpg"),
-            md5="16371574a10e0d10b88b807204c4f546",
-        ),
-    }
     base_johnson = (
         "https://raw.githubusercontent.com/jcjohnson/fast-neural-style/master/images/"
     )
@@ -218,7 +184,7 @@ def images() -> DownloadableImageCollection:
         ),
     }
     return DownloadableImageCollection(
-        {**texture_images, **content_images, **style_images}
+        {**content_images, **style_images}
     )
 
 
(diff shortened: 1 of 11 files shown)


pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -53,6 +53,10 @@ def content_transform(
     transforms_: List[transforms.Transform] = []
     if impl_params:
         if instance_norm:
+            # Necessary transformation, otherwise ValidRandomCrop will not work with too
+            # small images. This is therefore not in the reference implementation.
+            # TODO: ??
```
pmeier, line 58: How did they dealt with this?

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -99,6 +100,22 @@ def content_loss(
...
+    @staticmethod
+    def forward(self: Any, input_tensor: torch.Tensor, weight: float) -> torch.Tensor:  # type: ignore[override]
+        self.weight = weight
+        return input_tensor
+
+    @staticmethod
+    def backward(self: Any, grad_output: torch.Tensor) -> Tuple[torch.Tensor, Any]:  # type: ignore[override]
+        grad_input = grad_output.clone()
+        grad_input = grad_input / (torch.norm(grad_input, keepdim=True) + 1e-8)
+        return grad_input * self.weight, None
```
pmeier, line 116: Wouldn't this multiply the weights twice together with L180?

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -118,9 +135,14 @@ def __init__(
         # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L38
         # In the reference implementation the gram_matrix is only divided by the
         # batch_size.
-        normalize = not impl_params
         self.normalize_by_num_channels = impl_params
-        super().__init__(encoder, normalize=normalize, **gram_op_kwargs)
+        super().__init__(encoder, **gram_op_kwargs)
+
+        # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L56-L57
```
pmeier, line 141: Could you add a link that shows that `normalize` is set there?

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -118,9 +135,14 @@ def __init__(
         # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L38
         # In the reference implementation the gram_matrix is only divided by the
         # batch_size.
-        normalize = not impl_params
         self.normalize_by_num_channels = impl_params
-        super().__init__(encoder, normalize=normalize, **gram_op_kwargs)
+        super().__init__(encoder, **gram_op_kwargs)
```
pmeier, line 139: `normalize=True` by default. I think you mean to set `normalize=False`, didn't you?

pystiche_papers/ulyanov_et_al_2016/_modules.py
```diff
@@ -388,7 +388,7 @@ def conv_sequence(
         modules.append(("conv_seq", conv_seq))
         return SequentialWithOutChannels(OrderedDict(modules))
 
-    use_noise = not impl_params
+    use_noise = not impl_params or not instance_norm
```
pmeier, line 391: 1. Could you add links here?
2. Does that mean, that in the paper introducing `instance_norm` the additional noise is not used / mentioned at all?

pmeier (changes requested):
Some comments below. Could you split this into multiple PRs?

pystiche_papers/ulyanov_et_al_2016/_data.py
jbueltemeier: They simply take another image until the size fits.

pystiche_papers/ulyanov_et_al_2016/_loss.py
jbueltemeier: They also have a multiplication in the backward pass see [L60](https://github.com/DmitryUlyanov/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L60). However, in the solution of ProGamerGov is a further multiplication of the weight [L414](https://github.com/ProGamerGov/neural-style-pt/blob/cbcd023326a3487a2d75270ed1f3b3ddb4b72407/neural_style.py#L414), but I think that's just a typing there. But I'm just about to read the discussion of it.

pystiche_papers/ulyanov_et_al_2016/_modules.py
jbueltemeier: In the master branch the input is only filled with noise in `mode='texture'` otherwise it is ignored.

jbueltemeier: [L16-L18](https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/models/pyramid.lua#L16-L18)

pystiche_papers/ulyanov_et_al_2016/_loss.py
jbueltemeier: Yeah, you're right, I removed that by mistake. They are not using a normalisation in their gram calculation.

pystiche_papers/ulyanov_et_al_2016/_loss.py
pmeier: > They also have a multiplication in the backward pass see [L60](https://github.com/DmitryUlyanov/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/texture_loss.lua#L60).

In their implementation the forward and backward pass are independent of each other. That means, the multiplication in the forward pass does not affect the gradient at all. 

In our implementation we only need to define the forward pass and and the backward pass is added automatically. Since we have the multiplication in the forward pass, it is automatically included in the backward pass.

Thus, if you also include it into the scaling, it is performed twice.

pystiche_papers/ulyanov_et_al_2016/_modules.py
pmeier: For clarity I would prefer:

```suggestion
    use_noise = impl_params and not instance_norm
```

pystiche_papers/ulyanov_et_al_2016/_data.py
pmeier: In that case, we should do that too. I'll look into it how to do this.

jbueltemeier:
This PR has been divided into several smaller PRs:
[#250](https://github.com/pystiche/papers/pull/250) Manipulate gradients in ulyanov_et_al_2016
[#251](https://github.com/pystiche/papers/pull/251) Change Ulyanov stylization pretransformation
[#252 ](https://github.com/pystiche/papers/pull/252) Remove unused texture images in ulyanov_et_al_2016
[#253](https://github.com/pystiche/papers/pull/253) Add missing permalinks in ulyanov_et_al_2016

The handling of the image pre-processing of the content images has still been left open. This currently leads to a failure of the training. This will be addressed in a new PR.

PR #250, opened 2020-12-15
Manipulate gradients in ulyanov_et_al_2016

jbueltemeier:
Since in their implementation the forward pass and the backward pass are independent of each other, the backward pass must be manipulated in pytorch to reproduce the results. This includes the removal of gradients automatically created by pytorch, the addition of gradient normalization at certain points, and the addition of gradient normalization.


pmeier (changes requested):
I've installed and run the reference implementation of the authors to verify everything. I think your implementation was correct. Still, I suggested some changes that should make everything a little bit more clear.

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -56,7 +91,17 @@ def calculate_score(
         # uses reduction="mean" which also includes the batch_size. However, the
         # score is divided once more by the batch_size in the reference implementation.
         batch_size = image.extract_batch_size(input_repr)
-        return score / batch_size
+        # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/content_loss.lua#L23-L25
```
pmeier, line 94: [L162](https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_train.lua#L162) seems to only divide twice by the batch size for logging. In [L165](https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_train.lua#L165) you can see that the original loss is returned.

Following this I double checked with the reference implementation. IMO, the content loss behaves the same as a default `ops.FeatureReconstructionOperator`. If you are in doubt or have a test case where you think this is not the case, please let me know. Otherwise, I think we can remove this custom implementation completely.

pmeier (changes requested):

pystiche_papers/ulyanov_et_al_2016/_utils.py
```diff
@@ -34,9 +34,9 @@ def hyper_parameters(
             layers=("relu1_1", "relu2_1", "relu3_1", "relu4_1")
             if impl_params and instance_norm
             else ("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1"),
-            layer_weights="sum",
+            layer_weights=[1e3] * 4 if impl_params and instance_norm else [1e3] * 5,
             # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_train.lua#L23
-            score_weight=1e3 if impl_params and not instance_norm else 1e0,
+            score_weight=1e0 if impl_params and not instance_norm else 1e0,
```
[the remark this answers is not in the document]
jbueltemeier: The `score_weight` could actually be completely omitted here, since `score_weight=1e0` is the default value. 

Something also went wrong when transferring the correct `layer_weights`,  as this should be something like:
```python
layer_weights=1e3 if impl_params and not instance_norm else 1e0,
```
Since the `score_weight` is `1e3` in only one of the reference implementations.
 So I would suggest to use the code line above and set the length of the required Sequence[float] of the `MultiLayerEncodingOperator` in `style_loss`. Or what do you think?

pystiche_papers/ulyanov_et_al_2016/_loss.py
pmeier: I think it would be a good idea to split this into separate PR.

pystiche_papers/ulyanov_et_al_2016/_utils.py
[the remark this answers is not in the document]
pmeier: I'm sorry, but I don't understand your comment. Maybe there is some confusion about what `score_weight` and `layer_weights` are. Basically, each layer is weighted with `score_weight * layer_weights[i]` where `i` denotes the layer index. Thus

- `score_weight=1e3` and `layer_weights="sum"` (and thus `layer_weights=[1.0] * len(layers)`) and
- `score_weight=1e0` and `layer_weights=[1e3] * len(layers)`

are equal. The former has to advantage that it works for any number of layers whereas you need to specify it in the second case. Furthermore the former is conceptually more clear: Instead of the individual layers, the style loss is weighted. 

If that does not help could you please compile an example what you want to achieve with this?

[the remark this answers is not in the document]
jbueltemeier: I'm completely with you on the different weights. However, this will no longer work with the manipulated backward pass into the `GramOperator`, since the `score_weight`:
```python
score_weight=1e3 if impl_params and not instance_norm else 1e0
```
for each `GramOperator` is required here. If the `layer_weights` are `"sum"` the `score_weight` of these `GramOperator`s will always be `1e0`. This was also my intention why I changed the `layer_weights` here, because this way the operators will have the needed `score_weight`.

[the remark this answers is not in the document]
pmeier: I think I finally got it: I think you are saying that this is problematic since the backward pass of `ManipulateGradient` is **not** independent of `score_weight`. In the reference implementation the `score_weight` is `1e0` and the `layer_weight` is `1e3`, correct?

I've checked and indeed this is a problem. Good catch if you intended this and otherwise thank you for sending me on that journey :wink:

---

Here is what I would do about that:

You put 

```python
# https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/train.lua#L44
style_loss_layers = (
    ("relu1_1", "relu2_1", "relu3_1", "relu4_1")
    if impl_params and instance_norm
    else ("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1")
)
```

on the top of the function and within the hyper parameters definition you can do 

```python
style_loss = HyperParameters(
    layers=style_loss_layers,
    # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_train.lua#L23
    layer_weights=(
        [1e3 if impl_params and not instance_norm else 1e0] * len(style_loss_layers)
    ),
    score_weight=1e0,
)
```

You should also include a short comment here why we need to go with this approach rather than the usual one.

pystiche_papers/ulyanov_et_al_2016/_loss.py
```diff
@@ -17,6 +18,36 @@
...
+class RemoveBatchSizeDivisionGradient(torch.autograd.Function):
+    @staticmethod
+    def forward(self: Any, input_tensor: torch.Tensor, batch_size: int) -> torch.Tensor:  # type: ignore[override]
+        self.batch_size = batch_size
+        return input_tensor
+
+    @staticmethod
+    def backward(self: Any, grad_output: torch.Tensor) -> Tuple[torch.Tensor, Any]:  # type: ignore[override]
+        grad_input = grad_output.clone()
+        return grad_input * self.batch_size, None
```
pmeier, line 30: This seems to be not used anymore. Please revert it.

pystiche_papers/ulyanov_et_al_2016/_utils.py
```diff
@@ -31,12 +37,13 @@ def hyper_parameters(
...
         style_loss=HyperParameters(
             # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/train.lua#L44
-            layers=("relu1_1", "relu2_1", "relu3_1", "relu4_1")
-            if impl_params and instance_norm
-            else ("relu1_1", "relu2_1", "relu3_1", "relu4_1", "relu5_1"),
-            layer_weights="sum",
+            layers=style_loss_layers,
             # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_train.lua#L23
-            score_weight=1e3 if impl_params and not instance_norm else 1e0,
+            layer_weights=(
```
pmeier, line 42: Please include a comment why we do not use `score_weight` for this.

tests/unit/ulyanov_et_al_2016/test_loss.py
```diff
@@ -68,10 +68,8 @@ def test_GramOperator(
         normalize,
     ) in configs:
         with subtests.test(impl_params=impl_params):
-            target_repr = pystiche.gram_matrix(
-                encoder(target_image), normalize=normalize
-            )
-            input_repr = pystiche.gram_matrix(encoder(input_image), normalize=normalize)
+            target_repr = pystiche.gram_matrix(encoder(target_image), normalize=True)
```
pmeier, line 71: Since `normalize=True` is fixed, there is no need to include it in the loop variables (L68).

pmeier (changes requested):
In addition to the comments below, could you please revert all changes made to the custom `FeatureReconstructionOperator` as well as the `content_loss`?

pmeier (approved):
Thanks a lot!

PR #251, opened 2020-12-17
Change Ulyanov stylization pretransformation

jbueltemeier:


pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -131,6 +131,7 @@ def stylization(
     transformer: Union[nn.Module, str],
     impl_params: bool = True,
     instance_norm: bool = False,
+    edge_size: int = 256,
```
pmeier, line 134: Shouldn't this depend on `impl_params`? By default, I don't think the stylization should resize the image.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -144,6 +145,10 @@ def stylization(
         instance_norm: Switch the behavior and hyper-parameters between both
             publications of the original authors. For details see
             :ref:`here <ulyanov_et_al_2016-instance_norm>`.
+        edge_size: Size of the image to which the image should be resized before
+            transformation.
```
pmeier, line 149:
```suggestion
```

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -152,20 +157,25 @@ def stylization(
         transformer = _transformer(
             style=style, impl_params=impl_params, instance_norm=instance_norm,
         )
-        if instance_norm or not impl_params:
-            # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32
-            transformer = transformer.eval()
+    if instance_norm or not impl_params:
+        # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32
+        transformer = transformer.eval()
```
pmeier, line 162: Please revert this since it is discussed in #256.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -152,20 +157,25 @@ def stylization(
...
-        content_transform = _content_transform(
-            impl_params=impl_params, instance_norm=instance_norm
-        )
-        content_transform = content_transform.to(device)
-        input_image = content_transform(input_image)
+        if impl_params:
+            if hyper_parameters is None:
+                hyper_parameters = _hyper_parameters(
+                    impl_params=impl_params, instance_norm=instance_norm
+                )
```
pmeier, line 173: We should always do this near the top of the function even if we only use it here. This makes it consistent with the other functions for example `training`.

pmeier (changes requested):

pmeier (approved):

PR #256, opened 2020-12-18
Fix small trifles in ulyanov_et_al_2016

jbueltemeier:


pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -152,9 +154,9 @@ def stylization(
         transformer = _transformer(
             style=style, impl_params=impl_params, instance_norm=instance_norm,
         )
-        if instance_norm or not impl_params:
-            # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32
-            transformer = transformer.eval()
+    if instance_norm or not impl_params:
+        # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32
+        transformer = transformer.eval()
```
pmeier, line 159: I think the intention was that we only switch to evaluation mode if no transformer is passed explicitly. Thoughts?

jbueltemeier: But should this not be done regardless of whether a transformer is passed or a pre-trained one is used, if in their stylization function the transformer is set to evaluation mode. [L32](https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32)

pmeier: The issue I see is that we would change transformer in place without notice. IMO we should stick to only make this change if we create the transformer in here.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -152,9 +154,9 @@ def stylization(
         transformer = _transformer(
             style=style, impl_params=impl_params, instance_norm=instance_norm,
         )
-        if instance_norm or not impl_params:
-            # https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/test.lua#L32
-            transformer = transformer.eval()
+    if instance_norm or not impl_params:
```
pmeier, line 157: I think we should change the conditional to make it more clear:

```suggestion
    if not (impl_params and not instance_norm):
```

Although they are the same the suggested version makes it clear that there is only a single case where we don't switch to evaluation mode.
Thoughts?

pmeier (approved):

PR #264, opened 2022-02-22
upgrade gatys_et_al_2017

pmeier:
@jbueltemeier Same deal as before. If you see something obvious, feel free to comment. Otherwise, we can also do it later.


jbueltemeier (commented):
In the `gatys_et_al_2017` files there are still some references to `gatys_ecker_bethge_2016`. Especially in the comments on `multi_layer_encoder` and `hyperparameters`. This should be changed.

pmeier:
Could you mark one such instance?

pmeier:
Even better :heart:

PR #265, opened 2022-02-24
upgrade li_wand_2016

pmeier:


jbueltemeier (commented):
There are still regular expressions for the old `operator` in `._loss.py`.

PR #267, opened 2022-03-02
Correct Gatys ecker bethge 2016 Figure 3 bug

jbueltemeier:
Currently, the same style layers are always used in Figure 3 instead of the layers to be looped over. This PR corrects this by using the correct layers for the nst. Additionally, the layers from hyperparameter are used to switch the mapping between impl_param and the paper.


pmeier (approved):
Thanks!

Issue #275, opened 2022-03-12
Adjusting the output filenames depending on impl_params


jbueltemeier:
There are currently scripts to create the images for the parameters from the reference implementation as well as for the paper. However, the file name is currently the same for both scripts. Therefore, I would suggest that similar to `johnson_et_al_2016`, the name is adjusted depending on `impl_params` so that the files are not overwritten when both variants are created.

Example: `gatys_ecker_betghe_2016`
Current: "_fig_2__d.jpg_"
New: "_fig_2__d.jpg_" if `impl_params` else "_fig_2__d__paper.jpg_"

This would affect the scripts of:
`gatys_ecker_betghe_2016`
`gatys_et_al_2017`
`li_wand_2016`

What do you think @pmeier ?

pmeier:
SGTM, go for it!

PR #276, opened 2022-03-13
Change output filenames in replication scripts

jbueltemeier:
As mentioned in [#275](https://github.com/pystiche/papers/issues/275) the output file names are not yet ideal. This PR addresses this problem by making the file names more consistent. They now essentially consist of three components. 

1. The name of the replication.
2. Additional parameters that are important for the name (figure labels, hyperparameters, ...).
3. An optional flag indicating whether the `impl_params` are used.

Example of current filename:
"_gatys_ecker_betghe_2016__fig_2__B__impl_params.jpg_"

### DIFF
--- a/pystiche_papers/utils/misc.py
+++ b/pystiche_papers/utils/misc.py
@@ -40,6 +40,7 @@ __all__ = [
     "str_to_bool",
     "load_urls_from_csv",
     "select_url_from_csv",
+    "make_output_filename",
 ]
 
 
@@ -238,3 +239,9 @@ def select_url_from_csv(
             f"{fieldname}: {value}" for fieldname, value in zip(fieldnames, config)
         )
         raise RuntimeError(msg) from error
+
+
+def make_output_filename(name_parts: List[str], impl_params: bool = True, extension: str = ".jpg") -> str:
+    if impl_params:
+        name_parts.append("impl_params")
+    return "__".join(name_parts) + extension
--- a/pystiche_papers/utils/misc.py
+++ b/pystiche_papers/utils/misc.py
@@ -242,14 +242,8 @@ def select_url_from_csv(
 
 
 def make_output_filename(
-    name_parts: List[str],
-    impl_params: bool = True,
-    instance_norm: bool = False,
+    *parts: str,
     extension: str = ".jpg",
+    **flags: bool,
 ) -> str:
-    if impl_params:
-        name_parts.append("impl_params")
-
-    if instance_norm:
-        name_parts.append("instance_norm")
-    return "__".join(name_parts) + extension
+    return "__".join([*parts, *[name for name, flag in flags.items() if flag]]) + extension
--- a/pystiche_papers/utils/misc.py
+++ b/pystiche_papers/utils/misc.py
@@ -241,7 +241,9 @@ def select_url_from_csv(
         raise RuntimeError(msg) from error
 
 
-def make_output_filename(name_parts: List[str], impl_params: bool = True, extension: str = ".jpg") -> str:
+def make_output_filename(
+    name_parts: List[str], impl_params: bool = True, extension: str = ".jpg"
+) -> str:
     if impl_params:
         name_parts.append("impl_params")
     return "__".join(name_parts) + extension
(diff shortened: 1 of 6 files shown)


pystiche_papers/utils/misc.py
```diff
@@ -238,3 +239,17 @@ def select_url_from_csv(
...
+    impl_params: bool = True,
+    instance_norm: bool = False,
+    extension: str = ".jpg",
+) -> str:
+    if impl_params:
+        name_parts.append("impl_params")
+
+    if instance_norm:
+        name_parts.append("instance_norm")
+    return "__".join(name_parts) + extension
```
pmeier, line 255:
```suggestion
def make_output_filename(
    *parts: str,
    extension: str = ".jpg",
    **flags: bool,
) -> str:
    return "__".join([*parts, *[name for name, flag in flags.items() if flag]]) + extension
```

replication/gatys_ecker_bethge_2016/main.py
```diff
@@ -51,10 +51,13 @@ def __init__(self, label, image, score_weight):
             impl_params=args.impl_params,
             hyper_parameters=hyper_parameters,
         )
-
+        filename = utils.make_output_filename(
+            ["gatys_ecker_betghe_2016", "fig_2", style_image.label],
```
pmeier, line 55: With above's suggestion you only need to remove the list:

```suggestion
            "gatys_ecker_betghe_2016", "fig_2", style_image.label,
```

Of course this needs to be applied to all calls.

tests/unit/utils/test_misc.py
```diff
@@ -280,3 +280,24 @@ def test_select_url_from_csv_converter(urls_csv):
         file, config, converters={fieldname: utils.str_to_bool}
     )
     assert actual_url == expected_url
+
+
+def test_make_output_filename():
```
pmeier, line 285: We need to update these, but let's fix the other issues first.

pmeier (changes requested):
As discussed in today's meeting.

replication/gatys_ecker_bethge_2016/main.py
```diff
@@ -142,5 +152,5 @@ def process_dir(dir):
 if __name__ == "__main__":
     args = parse_input()
 
-    # figure_2(args)
+    figure_2(args)
```
pmeier, line 155: Good catch!

replication/gatys_et_al_2017/main.py
```diff
@@ -30,8 +30,11 @@ def figure_2_d(content_image, style_image):
             style_image,
             impl_params=args.impl_params,
         )
+        filename = make_output_filename(
+            "gatys_et_al_2017", "fig_2", "d", impl_params=args.impl_params
```
pmeier, line 34: With the extra comma at the end, `black` explodes this, which makes it easier to read.

```suggestion
            "gatys_et_al_2017", 
            "fig_2", 
            "d", 
            impl_params=args.impl_params,
```

Please apply this to all occurrences below.

tests/unit/utils/test_misc.py
```diff
@@ -280,3 +280,24 @@ def test_select_url_from_csv_converter(urls_csv):
...
+    name_parts = ["A", "B", "C"]
+    actual_filename = utils.make_output_filename(*name_parts)
+    expected_filename = "A__B__C__impl_params.jpg"
+    assert actual_filename == expected_filename
```
pmeier, line 289: You can use parametrization here

```suggestion
@pytest.mark.parametrize(
    ("args", "kwargs", "expected"),
    [
        (("A", "B", "C"), dict(), "A__B__C.jpg"),
        (("A", "B", "C"), dict(extension=".png"), "A__B__C.png"),
        (("A", "B", "C"), dict(impl_params=True), "A__B__C__impl_params.jpg"),
        (("A", "B", "C"), dict(impl_params=False), "A__B__C.jpg"),
    ],
)
def test_make_output_filename(args, kwargs, expected):
    assert utils.make_output_filename(*args, **kwargs) == expected
```

This is equivalent to the test you have written, but it is a lot simply to add another test case if we need it.

pmeier (approved):
Thanks!

PR #283, opened 2022-03-21
fix image sources for gatys 2017 et al

### LINKED ISSUE #281: Gatys 2017 content image download is broken
The image 

https://github.com/pystiche/papers/blob/36213e573fa98f2ceece9ba18b0899d19e6ab760/pystiche_papers/gatys_et_al_2017/_data.py#L33-L37

is no longer accessible. The [shop link](https://de.aliexpress.com/item/1705231003.html) still features it as thumbnail, but there is no direct link. Not sure what do to about it. I have the image locally, so in theory we could host it, but that is a copyright violation. Not sure if there are exemptions for research here. cc @jbueltemeier

pmeier:
Fixes #281 and reverts #282.

### DIFF
--- a/pystiche_papers/gatys_et_al_2017/_data.py
+++ b/pystiche_papers/gatys_et_al_2017/_data.py
@@ -1,101 +1,119 @@
+from typing import Optional
+
 from pystiche import data
 
 __all__ = ["images"]
 
 
+def license(original: Optional[str] = None) -> str:
+    license = (
+        "The image is part of a repository that is published for academic and "
+        "non-commercial use only "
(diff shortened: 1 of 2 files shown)


jbueltemeier (approved):

jbueltemeier:
Just a note. Can the `license` function in `pystiche_papers/utils`? I think we could also use it for other replications where we have used the images from the repository.
Like, for example: 
`johnson_et_al_2016`
`ulyanov_et_al_2016`

pmeier:
Sure. Would you mind sending a follow-up PR for that?

PR #285, opened 2022-03-24
Fix image sources

jbueltemeier:
In [#283](https://github.com/pystiche/papers/pull/283), a licence information for the images of `gatys_et_al_2017` from the original repository has been introduced. This information is generally useful for replications in which the images from the original repository are used. 

For this reason, this PR integrates this licence information for all current replications.

### DIFF
--- a/pystiche_papers/gatys_et_al_2017/_data.py
+++ b/pystiche_papers/gatys_et_al_2017/_data.py
@@ -14,9 +14,8 @@ def license(original: Optional[str] = None) -> str:
     if original:
         license_text = (
             f"{license_text} The original was probably downloaded from {original}. "
-            f"Proceed at your own risk."
         )
-    return license_text
+    return f"{license_text} Proceed at your own risk."
 
 
 def images() -> data.DownloadableImageCollection:
--- a/pystiche_papers/gatys_et_al_2017/_data.py
+++ b/pystiche_papers/gatys_et_al_2017/_data.py
@@ -1,7 +1,6 @@
 from typing import Optional
 
 from pystiche import data
-from pystiche_papers.utils import license
 
 __all__ = ["images"]
 
@@ -12,7 +11,12 @@ def license_info(original: Optional[str] = None) -> str:
         "non-commercial use only "
         "(https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/README.md?plain=1#L48)."
     )
-    return license(license_text, original=original)
+    if original:
+        license_text = (
+            f"{license_text} The original was probably downloaded from {original}. "
+            f"Proceed at your own risk."
+        )
+    return license_text
 
 
 def images() -> data.DownloadableImageCollection:
@@ -47,7 +51,7 @@ def images() -> data.DownloadableImageCollection:
             "watertown": data.DownloadableImage(
                 "https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/Images/ControlPaper/fig2_style1.jpg?raw=true",
                 file="watertown.jpg",
-                license=license("https://de.aliexpress.com/item/1705231003.html"),
+                license=license_info("https://de.aliexpress.com/item/1705231003.html"),
                 md5="4cc98a503da5ce6eab0649b09fd3cf77",
                 guides=data.DownloadableImageCollection(
                     {
--- a/pystiche_papers/gatys_et_al_2017/_data.py
+++ b/pystiche_papers/gatys_et_al_2017/_data.py
@@ -6,16 +6,14 @@ __all__ = ["images"]
 
 
 def license(original: Optional[str] = None) -> str:
-    license_text = (
+    license = (
         "The image is part of a repository that is published for academic and "
         "non-commercial use only "
         "(https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/README.md?plain=1#L48)."
     )
     if original:
-        license_text = (
-            f"{license_text} The original was probably downloaded from {original}. "
-        )
-    return f"{license_text} Proceed at your own risk."
+        license = f"{license} The original was probably downloaded from {original}. "
+    return f"{license} Proceed at your own risk."
 
 
 def images() -> data.DownloadableImageCollection:
@@ -25,7 +23,7 @@ def images() -> data.DownloadableImageCollection:
                 "https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/Images/ControlPaper/fig2_content.jpg?raw=true",
                 file="house_concept_tillamook.jpg",
                 license=license(
-                    "https://associateddesigns.com/sites/default/files/plan_images/main/craftsman_house_plan_tillamook_30-519-picart.jpg",
+                    "https://associateddesigns.com/sites/default/files/plan_images/main/craftsman_house_plan_tillamook_30-519-picart.jpg"
                 ),
                 md5="5629bf7b24a7c98db2580ec2a8d784e9",
                 guides=data.DownloadableImageCollection(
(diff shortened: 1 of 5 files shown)


pystiche_papers/utils/misc.py
```diff
@@ -238,3 +239,13 @@ def select_url_from_csv(
             f"{fieldname}: {value}" for fieldname, value in zip(fieldnames, config)
         )
         raise RuntimeError(msg) from error
+
+
+def license(license: str, original: Optional[str] = None) -> str:
```
pmeier, line 244: So this thing does nothing if `original` is not passed? What is the point to use it in that case?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -69,6 +73,22 @@ def forward(self, image: torch.Tensor) -> torch.Tensor:
         )
 
 
+def license_info(original: Optional[str] = None, repository: str = "ulyanov") -> str:
```
pmeier, line 76: If I'm not mistaken, `original` is never passed in this whole file. Given that `license` is a no-op in this case, why do we want to call it in the first place?

pmeier (commented):
Not sure I understand why we want to make the change. IIUC, this function is not used directly anywhere, but always wrapped inside another function. Can't we just simply add the extra sentence manually in there?

pystiche_papers/gatys_et_al_2017/_data.py
```diff
@@ -5,19 +5,18 @@
 __all__ = ["images"]
 
 
-def license(original: Optional[str] = None) -> str:
-    license = (
+def license_info(original: Optional[str] = None) -> str:
```
pmeier, line 8: Are there any changes here besides the name? If so, could you please revert?

pystiche_papers/johnson_alahi_li_2016/_data.py
```diff
@@ -30,6 +30,15 @@
 ]
 
 
+def license_info() -> str:
+    license_text = (
+        "The image is part of a repository that is published for personal and "
+        "research use only "
+        "(https://github.com/jcjohnson/fast-neural-style/blob/master/README.md#license)."
+    )
+    return license_text
```
pmeier, line 39: Since this functions takes no inputs and always returns the same output, can't we make it a constant?

```suggestion
LICENSE = (
    "The image is part of a repository that is published for personal and "
    "research use only "
    "(https://github.com/jcjohnson/fast-neural-style/blob/master/README.md#license)."
    )
```

pystiche_papers/li_wand_2016/_data.py
```diff
@@ -70,6 +70,20 @@ def extra_repr(self) -> str:
...
+def license_info(original: Optional[str] = None) -> str:
+    license_text = (
+        "The image is part of a repository that is published under the MIT License "
+        "(MIT) "
+        "(https://github.com/chuanli11/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/License#L1)."
+    )
+    if original:
+        license_text = (
+            f"{license_text} The original was probably downloaded from {original}. "
+            f"Proceed at your own risk."
```
pmeier, line 82: Should we have this disclaimer also for Gatys 2017?

pystiche_papers/li_wand_2016/_data.py
```diff
@@ -70,6 +70,20 @@ def extra_repr(self) -> str:
         return f"size={self.size}"
 
 
+def license_info(original: Optional[str] = None) -> str:
```
pmeier, line 73: Could we name this the same as Gatys 2017?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -69,6 +72,22 @@ def forward(self, image: torch.Tensor) -> torch.Tensor:
         )
 
 
+def license_info(repository: str = "ulyanov") -> str:
```
pmeier, line 75: Same as above for Johnson. Can't we have two constants like `LICENSE_JOHNSON` and `LICENSE_ULYANOV`?

pystiche_papers/li_wand_2016/_data.py
jbueltemeier: We have integrated this disclaimer for Gatys 2017, but you probably mean ulyanov and johnson. But this is only added if original is given. I would say that it would make sense to integrate this everywhere at the end of the licence by default.

jbueltemeier:
```python
def license(original: Optional[str] = None) -> str:
    license_text = (
        "The image is part of a repository that is published under the MIT License "
        "(MIT) "
        "(https://github.com/chuanli11/CNNMRF/blob/fddcf4d01e2a6ce201059d8bc38597f74a09ba3f/License#L1)."
    )
    if original:
        license_text = (
            f"{license_text} The original was probably downloaded from {original}. "
        )
    return f"{license_text} Proceed at your own risk."
```

pystiche_papers/gatys_et_al_2017/_data.py
```diff
@@ -6,18 +6,17 @@
 
 
 def license(original: Optional[str] = None) -> str:
-    license = (
+    license_text = (
```
pmeier, line 9: Could you also revert this change?

pystiche_papers/gatys_et_al_2017/_data.py
```diff
@@ -27,7 +26,7 @@ def images() -> data.DownloadableImageCollection:
                 "https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/Images/ControlPaper/fig2_content.jpg?raw=true",
                 file="house_concept_tillamook.jpg",
                 license=license(
-                    "https://associateddesigns.com/sites/default/files/plan_images/main/craftsman_house_plan_tillamook_30-519-picart.jpg"
+                    "https://associateddesigns.com/sites/default/files/plan_images/main/craftsman_house_plan_tillamook_30-519-picart.jpg",
```
pmeier, line 29: Was the extra comma added by you or by the code formatter? Please remove it and see if `black` puts it back. If it doesn't we can keep it that way.

pystiche_papers/gatys_et_al_2017/_data.py
```diff
@@ -77,7 +76,7 @@ def images() -> data.DownloadableImageCollection:
                 "https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/Images/ControlPaper/fig2_style2.jpg?raw=true",
                 file="wheat_field.jpg",
                 license=license(
-                    "https://upload.wikimedia.org/wikipedia/commons/thumb/c/ce/Wheat-Field-with-Cypresses-%281889%29-Vincent-van-Gogh-Met.jpg/1920px-Wheat-Field-with-Cypresses-%281889%29-Vincent-van-Gogh-Met.jpg"
+                    "https://upload.wikimedia.org/wikipedia/commons/thumb/c/ce/Wheat-Field-with-Cypresses-%281889%29-Vincent-van-Gogh-Met.jpg/1920px-Wheat-Field-with-Cypresses-%281889%29-Vincent-van-Gogh-Met.jpg",
```
pmeier, line 79: Same as above.

pystiche_papers/gatys_et_al_2017/_data.py
```diff
@@ -103,15 +102,15 @@ def images() -> data.DownloadableImageCollection:
                 "https://github.com/leongatys/NeuralImageSynthesis/blob/cced0b978fe603569033b2c7f04460839e4d82c4/Images/ControlPaper/fig3_content.jpg?raw=true",
                 file="schultenhof.jpg",
                 license=license(
-                    "https://upload.wikimedia.org/wikipedia/commons/8/82/Schultenhof_Mettingen_Bauerngarten_8.jpg"
+                    "https://upload.wikimedia.org/wikipedia/commons/8/82/Schultenhof_Mettingen_Bauerngarten_8.jpg",
```
pmeier, line 105: Same as above. If this was done in response to me saying "exploding things makes them easier to read", we are slightly on the wrong path. The statement is true if you have 2+ input arguments. Since we only have one here, there is no need to "explode".

pystiche_papers/johnson_alahi_li_2016/_data.py
```diff
@@ -114,45 +120,62 @@ def images() -> DownloadableImageCollection:
     content_root = urljoin(root, "content/")
     content_images = {
         "chicago": DownloadableImage(
-            urljoin(content_root, "chicago.jpg"), md5="16ea186230a8a5131b224ddde01d0dd5"
+            urljoin(content_root, "chicago.jpg"),
+            license=LICENSE,
```
pmeier, line 124: Nice! Didn't realize we missed the licenses before.

pystiche_papers/li_wand_2016/_data.py
```diff
@@ -70,6 +70,20 @@ def extra_repr(self) -> str:
         return f"size={self.size}"
 
 
+def license(original: Optional[str] = None) -> str:
+    license_text = (
```
pmeier, line 74: I think the variable is also called `license` in Gatys 2017.

pystiche_papers/li_wand_2016/_data.py
```diff
@@ -128,7 +139,7 @@ def images() -> data.DownloadableImageCollection:
             title="Blue Bottle",
             author="Christopher Michel (cmichel67)",
             date="02.09.2014",
-            license=data.NoLicense(),
+            license=license("https://www.flickr.com/photos/cmichel67/15112861945"),
```
pmeier, line 142: Are we only using this function once? If so, there is no need for it. Let's just put the message here.

pmeier (approved):
One small nit left. Otherwise LGTM!

pmeier (approved):
Thanks!

PR #77, opened 2020-07-22
Tests li wand 2016 loss

jbueltemeier:

PR #286, opened 2022-03-28
Skipping DataLoader in ulyanov_et_al_2016

jbueltemeier:
A current problem with the training in `ulyanov_et_al_2016` is that it causes an error if the images in the dataset are too small. The error is caused by trying to cropping to a larger image size. As mentioned in [#248](https://github.com/pystiche/papers/pull/248), the authors of the original publication have solved this by simply using a different image (see [L91-L100](https://github.com/pmeier/texture_nets/blob/aad2cc6f8a998fedc77b64bdcfe1e2884aa0fb3e/dataloader.lua#L91-L100)). 

This PR solves this problem by using a `SkipDataLoader` that uses the next image in case of an error. This is a minimal solution, which should be enough for this replication. Or do you have an idea with which this could be done better @pmeier?

As a side note, this affects maybe 100-200 images out of the nearly 40k images in the MSCOCO dataset.


pmeier:
What about adding a 

```py
import torch.utils.data


class SkipSmallIterableDataset(torch.utils.data.IterableDataset):
    def __init__(self, dataset, min_size):
        self.dataset = dataset
        self.min_size = min_size

    def __iter__(self):
        for sample in self.dataset:
            if sample >= self.min_size:
                yield sample


dataset = SkipSmallIterableDataset([6, 5, 4, 3, 2, 1, 4], min_size=4)

for sample in torch.utils.data.DataLoader(dataset, batch_size=1):
    print(sample)
```

```
tensor([6])
tensor([5])
tensor([4])
tensor([4])
```

jbueltemeier:
I have also thought about this, but have not found a simple solution in combination with a `BatchSampler`, which in the case of a small image yields the next index.

In addition, in at least one case, the dataset must be iterated over more than once (50k iteration).

pmeier:
Could you compile a minimal example with all the parts that we need so we have a clearer picture. If my implementation does not work, I'm happy to adapt yours. It is complicated though. Hence, I'm looking for easier alternatives.

jbueltemeier:
I am aware that my solution is not optimal. I just don't know how to change the `__getitem__` method and the `idx`.

Here is a minimal example where currently a `RuntimeError` is issued if the input is too small:
```py
import torch.utils.data
from pystiche.image import transforms
from pystiche_papers.data.utils import FiniteCycleBatchSampler


class SmallImageError(transforms.Transform):
    def __init__(self, min_size):
        super().__init__()
        self.min_size = min_size

    def forward(self, input_image: torch.Tensor) -> torch.Tensor:
        if input_image >= self.min_size:
            return input_image
        msg = "Small image error!"
        raise RuntimeError(msg)


class SkipSmallIterableDataset(torch.utils.data.Dataset):
    def __init__(self, dataset, min_size, transform):
        self.dataset = dataset
        self.min_size = min_size
        self.transform = transform

    def __len__(self) -> int:
        return len(self.dataset)

    def __getitem__(self, idx: int):
        data = self.dataset[idx]
        if self.transform:
            data = self.transform(data)
        return data


min_size = 2
transform = transforms.ComposedTransform(SmallImageError(min_size=min_size))
dataset = SkipSmallIterableDataset([6, 5, 4, 3, 2, 1, 4], min_size=min_size, transform=transform)
batch_sampler = FiniteCycleBatchSampler(dataset, num_batches=10, batch_size=1)
for sample in torch.utils.data.DataLoader(dataset, batch_sampler=batch_sampler):
    print(sample)
```
Current output (with reduced Error message):
```py
tensor([6])
tensor([5])
tensor([4])
tensor([3])
tensor([2])
RuntimeError: Small image error!
```
The output that is needed:
```py
tensor([6])
tensor([5])
tensor([4])
tensor([3])
tensor([2])
tensor([4])
tensor([6])
tensor([5])
tensor([4])
```
The needed output is achieved with the `SkipDataLoader` in this PR.

pmeier:
Skipping is only possible with iterable datasets. Unfortunately, they don't work with samplers or batch samplers. Thus, I think it would be the easiest to implement a small custom dataset that integrates the functionality we have in ``

```py
import itertools

import torch.utils.data
from torch import nn


class SmallImageError(nn.Module):
    def __init__(self, min_size):
        super().__init__()
        self.min_size = min_size

    def forward(self, input_image: torch.Tensor) -> torch.Tensor:
        if input_image >= self.min_size:
            return input_image
        msg = "Small image error!"
        raise RuntimeError(msg)


class Dataset(torch.utils.data.IterableDataset):
    def __init__(self, dataset, min_size, transform, num_samples):
        self.dataset = dataset
        self.min_size = min_size
        self.transform = transform
        self.num_samples = num_samples

    def __iter__(self):
        dataset = itertools.cycle(self.dataset)
        num_samples = 0
        while num_samples < self.num_samples:
            sample = next(dataset)
            if sample >= self.min_size:
                yield transform(sample)
                num_samples += 1


min_size = 2
transform = SmallImageError(min_size=min_size)
dataset = Dataset(
    [6, 5, 4, 3, 2, 1, 4],
    min_size=min_size,
    transform=transform,
    # num_batches * batch_size
    num_samples=10 * 1,
)

for idx, sample in enumerate(torch.utils.data.DataLoader(dataset), 1):
    print(idx, sample)
```

```
1 tensor([6])
2 tensor([5])
3 tensor([4])
4 tensor([3])
5 tensor([2])
6 tensor([4])
7 tensor([6])
8 tensor([5])
9 tensor([4])
10 tensor([3])
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +188,97 @@ def images() -> DownloadableImageCollection:
     return DownloadableImageCollection({**content_images, **style_images})
 
 
+class SkipSmallIterableImageFolderDataset(torch.utils.data.IterableDataset):
```
pmeier, line 191: Not sure why you want to subclass here. Composition should be way easier. Why not simply do

```py
dataset = SkipSmallImagesDataset(ImageFolderDataset(root), transform=transform, ...)
```

where `SkipSmallImagesDataset` is `Dataset` from https://github.com/pystiche/papers/pull/286#issuecomment-1082832974?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -1,18 +1,19 @@
-from typing import List, Optional, Sized
+from typing import List, Optional
 from urllib.parse import urljoin
+import itertools
 
-from torch.utils.data import DataLoader, Dataset, Sampler
+import torch.utils.data
+from torch.utils.data import DataLoader, Dataset
```
pmeier, line 6: Why both? Either import everything you need by name or only import the namespace and access everything from there.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -21,8 +22,8 @@
     "content_transform",
     "style_transform",
     "images",
+    "SkipSmallIterableDataset",
```
pmeier, line 25: We shouldn't expose this as it shouldn't be used by users directly.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +187,71 @@ def images() -> DownloadableImageCollection:
...
+        dataset: torch.utils.data.Dataset,
+        min_size: int,
+        num_samples: int,
+        transform: transforms.Transform,
+    ):
+        self.dataset = dataset
+        self.min_size = min_size
+        self.num_samples = num_samples
+        self.transform = transform
+        self.data_samples = itertools.cycle(self.dataset)
```
pmeier, line 202: Why do this in the constructor? This means the dataset is not "reset" if whenever we call `iter` on it. For example:

```py
import itertools


class Dataset:
    def __init__(self):
        self.data_samples = itertools.cycle([1, 2, 3])

    def __iter__(self):
        yield from self.data_samples


dataset = Dataset()

for epoch in range(2):
    for idx, sample in enumerate(dataset):
        print(sample)
        if idx == 3:
            break
```

```
1
2
3
1
2
3
1
2
```

Compare that to

```py
class Dataset:
    def __init__(self):
        self.dataset = [1, 2, 3]

    def __iter__(self):
        yield from itertools.cycle(self.dataset)
```

```
1
2
3
1
1
2
3
1
```

Or is this behavior intended here?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +187,71 @@ def images() -> DownloadableImageCollection:
...
+
+    def __len__(self) -> int:
+        return self.num_samples
+
+    def __iter__(self) -> torch.Tensor:
+        num_samples = 0
+        while num_samples < self.num_samples:
+            sample = next(self.data_samples)
+            if all(size >= self.min_size for size in extract_image_size(sample)):
```
pmeier, line 211: Could you re-add the comment you had here before?

pystiche_papers/ulyanov_et_al_2016/_data.py
jbueltemeier: Yes, this is intended here so that the dataset is not reset after each epoch.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,74 @@ def images() -> DownloadableImageCollection:
     return DownloadableImageCollection({**content_images, **style_images})
 
 
+class SkipSmallIterableDataset(IterableDataset):
```
pmeier, line 188: After some more thought, lets just call this `Dataset` here

```suggestion
class Dataset(IterableDataset):
```

`SkipSmallIterableDataset` is only partially descriptive, since we also cycle through the data up to a given number of samples.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,74 @@ def images() -> DownloadableImageCollection:
...
+    def __init__(
+        self,
+        dataset: Dataset,
+        min_size: int,
```
pmeier, line 192: Let's make everything but the dataset keyword only. This makes the instantiation more readable.

```suggestion
        dataset: Dataset,
        *
        min_size: int,
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,74 @@ def images() -> DownloadableImageCollection:
...
+
 def dataset(
     root: str,
     impl_params: bool = True,
     instance_norm: bool = True,
     transform: Optional[transforms.Transform] = None,
-) -> ImageFolderDataset:
+    hyper_parameters: Optional[HyperParameters] = None,
+) -> SkipSmallIterableDataset:
     if transform is None:
```
pmeier, line 224: `content_transform()` also takes `hype_parameters`. Thus, we need to create it before and pass it on.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,74 @@ def images() -> DownloadableImageCollection:
...
-    return FiniteCycleBatchSampler(
-        data_source,
-        num_batches=hyper_parameters.batch_sampler.num_batches,
-        batch_size=hyper_parameters.batch_sampler.batch_size,
+    min_size = hyper_parameters.content_transform.edge_size
+    num_samples = (
+        hyper_parameters.sampler.num_samples * hyper_parameters.sampler.batch_size
+    )
+    return SkipSmallIterableDataset(
+        ImageFolderDataset(root), min_size, num_samples, transform
```
pmeier, line 237:
```suggestion
        ImageFolderDataset(root), 
        min_size=hyper_parameters.content_transform.edge_size, 
        num_samples=hyper_parameters.sampler.num_samples * hyper_parameters.sampler.batch_size, 
        transform=content_transform(impl_params=impl_params, instance_norm=instance_norm, hyper_parameters=hyper_parameters),
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,74 @@ def images() -> DownloadableImageCollection:
...
         )
-
-    return FiniteCycleBatchSampler(
-        data_source,
-        num_batches=hyper_parameters.batch_sampler.num_batches,
-        batch_size=hyper_parameters.batch_sampler.batch_size,
+    min_size = hyper_parameters.content_transform.edge_size
+    num_samples = (
+        hyper_parameters.sampler.num_samples * hyper_parameters.sampler.batch_size
+    )
```
pmeier, line 235: Something is off with the naming here. `num_samples` shouldn't be `num_samples` times anything. Maybe rename the one in the hyperparameters back to `num_batches`?

pystiche_papers/ulyanov_et_al_2016/_utils.py
```diff
@@ -62,10 +62,10 @@ def hyper_parameters(
             # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/src/descriptor_net.lua#L17
             else "bilinear",
         ),
-        batch_sampler=HyperParameters(
+        sampler=HyperParameters(
```
pmeier, line 65: Given that we don't have a `sampler` anymore, we should just remove this level to be able to do `hyper_parameters.num_samples`. The same applies to `batch_size`.

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -1,18 +1,18 @@
...
     DownloadableImage,
     DownloadableImageCollection,
     ExpiredCopyrightLicense,
-    ImageFolderDataset,
 )
-from pystiche.image import transforms
-from pystiche_papers.data.utils import FiniteCycleBatchSampler
+from pystiche.image import transforms, extract_image_size
 from pystiche_papers.utils import HyperParameters
+from pystiche.data import ImageFolderDataset
```
pmeier, line 15: Could you move this back to the import of the licenses above?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,78 @@ def images() -> DownloadableImageCollection:
...
+        self.dataset = dataset
+        self.min_size = min_size
+        self.num_samples = num_samples
+        self.transform = transform
+        self.data_samples = itertools.cycle(self.dataset)
+
+    def __len__(self) -> int:
+        return self.num_samples
+
+    def __iter__(self) -> Generator:
```
pmeier, line 206:
```suggestion
    def __iter__(self) -> Iterator:
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +185,78 @@ def images() -> DownloadableImageCollection:
...
+
+
 def dataset(
     root: str,
     impl_params: bool = True,
     instance_norm: bool = True,
     transform: Optional[transforms.Transform] = None,
-) -> ImageFolderDataset:
+    hyper_parameters: Optional[HyperParameters] = None,
+) -> SkipSmallIterableDataset:
```
pmeier, line 224:
```suggestion
) -> Dataset:
```

pmeier (commented):
These are some things the linter workflow complains about.

tests/unit/ulyanov_et_al_2016/test_data.py
```diff
@@ -106,47 +106,49 @@ def test_images_smoke():
...
 
     with subtests.test("content_transform"):
         assert isinstance(dataset.transform, type(paper.content_transform()))
 
+    with subtests.test("min_size"):
+        assert dataset.min_size == hyper_parameters.content_transform.edge_size
+
+    with subtests.test("num_samples"):
+        assert (
+            dataset.num_samples
```
pmeier, line 135:
```suggestion
    with subtests.test("length"):
        assert (
            len(dataset)
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -186,71 +186,76 @@ def images() -> DownloadableImageCollection:
...
+        *,
+        min_size: int,
+        num_samples: int,
+        transform: transforms.Transform,
+    ):
+        self.dataset = dataset
+        self.min_size = min_size
+        self.num_samples = num_samples
+        self.transform = transform
+        self.data_samples = itertools.cycle(self.dataset)
```
pmeier, line 202:
```suggestion
        self.data_samples = itertools.cycle(self.dataset)  # type:ignore [var-annotated, arg-type]
```

pmeier (commented):
For the doctest failure remove:

https://github.com/pystiche/papers/blob/c955e171a414c4d659a0dc0a8c7f1e07753d3bc3/docs/source/api/pystiche_papers.ulyanov_et_al_2016.rst?plain=1#L196

Download failure is unrelated. I'll fix in another PR.

pmeier:
I learned today, that `IterableDataset` needs special handling if used with more than one worker: https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset. Although we set `num_workers=0` by default, we cannot guarantee the users will do the same. Do you want me to fix this here, or should we add an issue and fix it later?

jbueltemeier:
That's interesting, thanks for the comment. Since this PR is only for this dataset, I think it makes sense to integrate it here. So feel free to do it here.

jbueltemeier:
I also learned something today (see this [Issue](https://github.com/pytorch/pytorch/issues/23900)):
> `itertools.cycle` attempts to save all outputs in order to re-cycle through them

This increases the memory with each iteration until the script crashes. Currently, however, `itertools.cycle` is required for the `Dataset` to cycle through the images several times. 

One solution would be from [#77](https://github.com/DeepLearnPhysics/lartpc_mlreco3d/pull/77):
```py
        def cycle(data_io):
            while True:
                for x in data_io:
                    yield x
        self.data_samples = iter(cycle(self.dataset)) 
```

Or do you see a better solution @pmeier?

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -199,7 +198,16 @@ def __init__(
...
         self.transform = transform
-        self.data_samples = itertools.cycle(self.dataset)  # type:ignore [var-annotated, arg-type]
+
+        # Use this function instead of itertools.cycle to avoid creating a memory leak.
+        # itertools.cycle attempts to save all outputs in order to re-cycle through them
+        # see https://github.com/DeepLearnPhysics/lartpc_mlreco3d/pull/77
+        def cycle(data_iter):
+            while True:
+                for x in data_iter:
+                    yield x
```
pmeier, line 208:
```suggestion
        # Like itertools.cycle but without caching
        def cycle(iterable):
            while True:
                for item in iterable:
                    yield item
```

pmeier (approved):
Thanks @jbueltemeier!

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -257,69 +255,83 @@ def images() -> DownloadableImageCollection:
...
 
 
+class Dataset(IterableDataset):
+    def __init__(
+        self,
+        dataset: Sized,
+        *,
+        min_size: int,
+        num_samples: int,
+        transform: Optional[nn.Module],
```
pmeier, line 265:
```suggestion
        transform: nn.Module,
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -257,69 +255,83 @@ def images() -> DownloadableImageCollection:
...
+    ):
+        self.dataset = dataset
+        self.min_size = min_size
+        self.num_samples = num_samples
+        self.transform = transform
+
+        # Like itertools.cycle but without caching
+        def cycle(iterable: Sized) -> Iterator:
```
pmeier, line 273:
```suggestion
        def cycle(iterable: Iterable) -> Iterator:
```

pystiche_papers/ulyanov_et_al_2016/_data.py
```diff
@@ -1,11 +1,11 @@
-from typing import cast, List, Optional, Sized, Tuple, Union
+from typing import List, Sized, Optional, Tuple, cast, Union, Iterator
```
pmeier, line 1:
```suggestion
from typing import List, Iterable, Optional, Tuple, cast, Union, Iterator
```

pmeier (commented):
Please fix the lint. When CI is happy feel free to merge.

Issue #294, opened 2022-09-30
Runtimeerror with single sample image in ulyanov_et_al_2016


jbueltemeier:
The following `RuntimeError` occurs when performing a stylisation with a transformer in `ulyanov_et_al_2016` with `impl_params=False`:

```python
RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 26 but got size 27 for tensor number 1 in the list.
```
The reason for this is the lack of image preprocessing, which prevents the image from being resized to a power of 2. This works well with most of the sample images, only with the image `"bird"` does the error occur. 

Since this error is to be expected due to the missing preprocessing (the image pyramid has 6 levels) and this only affects this one image in this replication, I would suggest to remove this image from the replication script and issue an appropriate warning at `impl_params=False` in the `stylisation` function.

Or what do you mean @pmeier.

pmeier:
IIUC, the images should have have side lengths, i.e. height and width, divisible by `2 ** 6 == 64`, correct? If yes, can't we simply resize the images to the nearest integer multiple of that? For example:

```py
image = ...
old_length = min(image.shape[-2:])
new_length = math.ceil(old_length / 64) * 64
resized_image = resize(image, new_length)
cropped_image = center_crop(resized_image, new_length)
new_height, new_width = cropped_image.shape[-2:]
assert new_height % 64 == 0
assert new_width % 64 == 0
```

That would enable people to use our stylization script with all images even their own without worrying about the size first. Since this is evaluation, we also don't need to worry about runtime costs.

pmeier:
Or maybe no even `center_crop`, but just crop to the nearest multiple. `johnson_alahi_li_2016` does something similar albeit at runtime:

https://github.com/pystiche/papers/blob/c5e075231d58fa1556a47a50d851a72f21ded443/pystiche_papers/johnson_alahi_li_2016/_data.py#L40

jbueltemeier:
Yes, that sounds like a good idea. I would also prefer to crop to the nearest multiple so as not to change the image unnecessarily much.

I will implement this.

PR #295, opened 2022-10-05
Ulyanov image stylization RuntimeError

jbueltemeier:
As described in [#294](https://github.com/pystiche/papers/issues/294), a RuntimeError currently occurs for images with a height or width that are not multiples of 64. This PR solves this problem by using an additional `TopLeftCropToMultiple`  for the case with `impl_params=False`.

### DIFF
--- a/pystiche_papers/ulyanov_et_al_2016/_nst.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_nst.py
@@ -8,7 +8,7 @@ from torchvision import transforms
 from pystiche import loss, misc, optim
 from pystiche_papers.utils import HyperParameters
 
-from ..utils import batch_up_image
+from ..utils import batch_up_image, TopLeftCropToMultiple
 from ._data import images as _images, style_transform as _style_transform
 from ._loss import perceptual_loss
 from ._modules import transformer as _transformer
@@ -170,7 +170,13 @@ def stylization(
             # https://github.com/pmeier/texture_nets/blob/b2097eccaec699039038970b191780f97c238816/stylization_process.lua#L30
             edge_size = hyper_parameters.content_transform.edge_size
             transform = transforms.Resize((edge_size, edge_size))
-            input_image = transform(input_image)
+        else:
+            transform = TopLeftCropToMultiple(multiple=64)
+
+        input_image = transform(input_image)
+        new_height, new_width = input_image.shape[-2:]
+        assert new_height % 64 == 0
+        assert new_width % 64 == 0
 
         output_image = transformer(input_image)
         output_image = postprocessor(output_image)
--- a/pystiche_papers/ulyanov_et_al_2016/_nst.py
+++ b/pystiche_papers/ulyanov_et_al_2016/_nst.py
@@ -171,6 +171,10 @@ def stylization(
             edge_size = hyper_parameters.content_transform.edge_size
             transform = transforms.Resize((edge_size, edge_size))
         else:
+            # No image pre-processing is described in the paper. However, images with a
+            # height or width that is not divisible by 64 will result in a RuntimeError.
+            # In order to be able to use the stylisation for all images, a cropping is
+            # carried out here.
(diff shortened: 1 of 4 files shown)


pmeier (commented):
Re-reading my comments, I think I need to walk back on https://github.com/pystiche/papers/issues/294#issuecomment-1263353343 as it is misleading. The problem I see is that 64 as multiple is pretty large and might crop a significant portion of the image. Thus, we should probably resize before we start cropping. Did you look at the source of the authors? What do they do before the stylization?

jbueltemeier:
The problem is that the authors do not specify in their paper (`impl_params=False`) how the images are preprocessed. They only show images in different sizes, with a height and width that are multiples of 64.

pmeier:
Could you have a look what the author is doing in the implementation?

jbueltemeier:
In the implementation, only a resize to the required image size is used. For the example images, the image size is 256x256. This results in a deformation of the image.

pmeier:
I'm not sure I understand. Could you link the parts you refer to?

pmeier:
As discussed offline:

1. Check if both side lengths are a multiple of 64. If yes, return.
2. Resize the image so that the smaller size is the next larger multiple of 64.
3. Center crop so both side lengths are a multiple of 64.

pystiche_papers/utils/transforms.py
```diff
@@ -4,7 +4,7 @@
 from pystiche import image
 
 __all__ = [
-    "OptionalGrayscaleToFakegrayscale",
```
pmeier, line 7: Could you revert that?

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -23,6 +24,26 @@
 __all__ = ["training", "stylization"]
 
 
+class OptionalResizeCenterCropToMultiple(nn.Module):
```
pmeier, line 27: Shouldn't this be in the `_data` module? I mean we have 

https://github.com/pystiche/papers/blob/2c3972838cb9b8998c86cd30fd549bb94b30d532/pystiche_papers/ulyanov_et_al_2016/_data.py#L84

Should we maybe rename that to `training_content_transform` and have a `stylization_content_transform`? Or maybe we don't need to rename the old one and just call this one `stylization_transform` since at this stage there is no longer a content or style image.

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -23,6 +24,26 @@
...
+    def __init__(self, multiple: int = 64):
+        super().__init__()
+        self.multiple = multiple
+
+    def forward(self, input_image: torch.Tensor) -> torch.Tensor:
+        old_height, old_width = input_image.shape[-2:]
+        if old_height % 64 == 0 and old_width % 64 == 0:
```
pmeier, line 34:
```suggestion
        if old_height % self.multiple == 0 and old_width % multiple == 0:
```

pystiche_papers/ulyanov_et_al_2016/_nst.py
```diff
@@ -23,6 +24,26 @@
...
+            return input_image
+
+        min_length = min([old_height, old_width])
+        new_length = math.ceil(min_length / self.multiple) * self.multiple
+        transforms_: List[nn.Module] = []
+        transforms_.append(transforms.Resize(new_length))
+        transforms_.append(transforms.CenterCrop(new_length))
+        transform = nn.Sequential(*transforms_)
+
+        return cast(torch.Tensor, transform(input_image))
```
pmeier, line 44: Instead of creating a transform object here and afterwards only calling it once, you can use the functional API directly:

```suggestion
        # move this to the other imports
        from torchvision.transforms import functional as F
        
        output_image = F.resize(input_image, new_length)
        output_image = F.center_crop(output_image, new_length)
        return output_image
```

pmeier (approved):
LGTM, thanks Julian!

PR #296, opened 2022-10-19
Add author info replication script in gatys_ecker_betghe_2016

jbueltemeier:
The replication of gatys_ecker_betghe_2016 shows clear differences in the results. For this reason, I have contacted the author. He provided helpful tips for improving the results. This information on the hyperparameters is integrated with this PR. The most important changes are:
1. the use of the activations after the ReLU function.
2. the use of `layer_weights` depending on the number of channels.
3. random initialisation of the image and more optimisation steps.

Due to the random initalisation, an exact replication is no longer possible, but the results show a clear improvement and a more similar stylisation to the original images.

### DIFF
--- a/replication/gatys_ecker_bethge_2016/main.py
+++ b/replication/gatys_ecker_bethge_2016/main.py
@@ -150,8 +150,9 @@ def figure_2_author_information(args):
 
     for style_image in style_images:
         print(f"Replicating Figure 2 {style_image.label} with author information")
-        # The ratio is between a larger and a smaller value, according to the
-        # author the larger value was 1e6 or 1e9 (1e9 dont work)
+        # The weight ratio from the paper is between a larger and a smaller value,
+        # according to the author the larger value was one of the discrete values 1e6 or
+        # 1e9 (1e9  leads to a Runtime Error)
         larger_value = 1e6
         hyper_parameters.style_loss.score_weight = larger_value
         hyper_parameters.content_loss.score_weight = (
@@ -161,7 +162,6 @@ def figure_2_author_information(args):
         output_image = paper.nst(
             content_image,
             style_image.image,
-            impl_params=True,
             hyper_parameters=hyper_parameters,
         )
 
--- a/replication/gatys_ecker_bethge_2016/main.py
+++ b/replication/gatys_ecker_bethge_2016/main.py
@@ -117,6 +117,67 @@ def figure_3(args):
         )
 
 
+def figure_2_author_information(args):
+    images = paper.images()
+    images.download(args.image_source_dir)
+
+    ####################author provided this information################################
+    # compute layer_weights and use relu output -> impl_params=True
+    hyper_parameters = paper.hyper_parameters(impl_params=True)
+
+    hyper_parameters.nst.num_steps = 2000
+    hyper_parameters.nst.starting_point = "random"
+    ####################################################################################
+    content_image = images["neckarfront"].read(
+        size=hyper_parameters.nst.image_size, device=args.device
+    )
+
+    class StyleImage:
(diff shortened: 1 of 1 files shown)


replication/gatys_ecker_bethge_2016/main.py
```diff
@@ -117,6 +117,67 @@ def figure_3(args):
...
+        StyleImage("D", images["the_scream"], 5e-3),
+        StyleImage("E", images["femme_nue_assise"], 5e-4),
+        StyleImage("F", images["composition_vii"], 5e-4),
+    )
+
+    for style_image in style_images:
+        print(f"Replicating Figure 2 {style_image.label} with author information")
+        # The ratio is between a larger and a smaller value, according to the
+        # author the larger value was 1e6 or 1e9 (1e9 dont work)
+        larger_value = 1e6
```
pmeier, line 155: So here is still uncertainty and the author also couldn't give us the data he used for producing the results in the paper? Could you detail what doesn't work for `1e9`?

replication/gatys_ecker_bethge_2016/main.py
```diff
@@ -117,6 +117,67 @@ def figure_3(args):
...
+        hyper_parameters.style_loss.score_weight = larger_value
+        hyper_parameters.content_loss.score_weight = (
+            style_image.weight_ratio * larger_value
+        )
+
+        output_image = paper.nst(
+            content_image,
+            style_image.image,
+            impl_params=True,
+            hyper_parameters=hyper_parameters,
```
pmeier, line 165: Does `impl_params=True` do anything if we pass `hyper_parameters` here?

pmeier (commented):
Thanks a lot for the effort Julian!

> He provided helpful tips for improving the results.

Just to be sure I understand correctly: with improvement you mean better alignment of our replicated results with what is reported in the paper, right? Because that is what we are striving for. General quality improvement of the images is not a goal here.

In general, change looks good. In the future we should probably look at removing `impl_params: bool` in favor of `config: GatysConfig` like

```py
class GatysConfig(enum.Enum):
    PAPER = enum.auto()
    REFERENCE_IMPLEMENTATION = enum.auto()
    AUTHOR = enum.auto()
```

This would make it much easier to deal with such a situation. This would also cover the `instance_norm: bool` thing that we have for Ulyanov.

replication/gatys_ecker_bethge_2016/main.py
pmeier: As discussed offline, the author provided these two *discrete* values. We'll recheck `1e9`, but otherwise `1e6` is the "right" one.

jbueltemeier: 1. I have looked at the commentary again and the ratio here refers to the weight ratio given in the paper. For this, there must be a larger value and a smaller value, and according to the author, the larger value (weight for `style_loss`) was the discrete value `1e6` or `1e9`.
2. The optimisation is unstable with a weighting of 1e9. The following error message appears regularly:
```python
File "/home/julian/github/pystiche_papers/.venv/lib/python3.7/site-packages/torch/autograd/__init__.py", line 175, in backward
    allow_unreachable=True, accumulate_grad=True)  # Calls into the C++ engine to run the backward pass
RuntimeError: Function 'ConvolutionBackward0' returned nan values in its 1th output.`
```

replication/gatys_ecker_bethge_2016/main.py
jbueltemeier: Thanks for the comment pmeier. I had a thinking error there and thought that this would activate the Relu output. But this is already done in the hyperparameters.

pmeier (approved):
Thanks!

